PR #1, opened 2015-05-11
Add basic playbook and a set of roles

enykeev:


enykeev:
Extract from StackStorm/st2workroom#27

Issue #2, opened 2015-08-11
General Plan


armab:
Ansible deployment repo definitely needs some polishing/testing/improvements for mass-use.

The big point is to polish & publish it on [Ansible galaxy](https://galaxy.ansible.com/) (public hub) to get some visibility in the community. Another good thing are real [CI Integration tests](https://github.com/StackStorm/ansible-st2/issues/15), for better maintainability.

TODO: 
#### Part 1
- [x] Check for minimum requirements like `memory` (`disk` space for Mongo?)
- [x] Version check before install (check if `13.5.99` version is available)
- [x] `stable`, `unstable` support (besides of numeric `0.12.1` version)
- [x] Vagrantfile
- [x] Refactor all variables, add `st2_` prefix (best practice for Ansible Galaxy)
- [x] Upgrade to higher st2 version as well as downgrade should work
- [x] Fix Upstart
- [x] Upstart `N` action-runners
- [x] Initial Docs 

---
- [x] Support [New packages](https://stackstorm.com/2016/05/03/road-to-packages/)
- [x] Add WebUI installation
- [x] ~Use external/3rd party Ansible Galaxy roles for PostgreSQL/Mongo/RabbitMQ~
- [x] Tag all tasks, split into logical groups
- [x] Idempotence (playbook re-run shouldn't change anything unless new st2 version is available)
- [x] CI tests for better maintainability (Test-kitchen?, CircleCI?)
- [x] Extend list of variables/options to configure
- [x] Ability to configure with existing/remote installations of Mongo, RabbitMQ, PostgreSQL
- [x] Publish on Ansible galaxy https://github.com/StackStorm/ansible-st2/issues/12
  - [x] Polish, Refactor, Follow best practices
  - [x] Distribution
    - [x] Ship it as single `stackstorm.stackstorm` https://galaxy.ansible.com/StackStorm/stackstorm/
    - [ ] Split into: `st2`, `st2mistral`, `st2web`, `st2chatops` independent roles (`v1.0` ??)
  - [x] Ansible Galaxy Docs, Examples
- [x] More distros to support (currently Ubuntu is supported only)

cognifloyd:
This has "Use external/3rd party Ansible Galaxy roles for PostgreSQL/Mongo/RabbitMQ" checked, but those changes have been removed.

armab:
@cognifloyd Updated, thanks.
[...]

PR #3, opened 2015-08-11
Small changes [RFR]

armab:
* [x] Check for minimum requirements like `memory`
* [x] Version check before install (check if `13.5.99` version is available)
* [x] `stable`, `unstable` support (besides of numeric `0.12.1` version)
* [x] Vagrantfile
* [ ] ~~Tag all tasks, split into logical groups~~ (next PR)
* [x] Refactor all variables, add `st2_` prefix (best practice for Ansible Galaxy)
* [x] Upgrade to higher st2 version as well as downgrade should work (it doesn't)
* [ ] ~~Add missing [PostgreSQL installation](https://github.com/StackStorm/st2/blob/b6b87aeb62c8301409b6b8fe16a6d4826560ff4f/tools/st2_deploy.sh#L280) for Mistral~~ (next PR)
* [x] Fix Upstart
* [x] Upstart `N` action-runners
* [ ] ~~Further verify/sync existing logic with [`st2_deploy.sh`](https://github.com/StackStorm/st2/blob/master/tools/st2_deploy.sh)~~ (next PR)
* [x] Initial Docs 

Part of #2


armab:
@jfryman @enykeev This PR is ready for review/merging.

Initial docs: https://github.com/armab/ansible-st2/blob/small-fixes/README.md

jfryman:
Did this get pushed to Ansible Galaxy, or is that after some more improvements?

armab:
Yeah, it's still far away to be ready for Ansible Galaxy.

I need to remove existing roles (MySQL, Mongo, RabbitMQ, PostgreSQL), instead of that reuse 3rd party Ansible Galaxy roles. As a result there will be only `st2` role with dependencies.
Many more improvements/fixes from [that checklist](https://github.com/StackStorm/ansible-st2/issues/2) needed, including ci/tests to make it good enough for Galaxy.

Here are just minimal changes which I'm sure won't change during time, like renamed variable names. I'm moving to that point (Ansible Galaxy and Stability) in small steps.

jfryman:
@armab good deal. Thanks for the info. I assumed as much, but you know what happens when one assumes. :grin:

armab:
No problem with that.
I should provide more clear descriptions.

And we're on the same wave about Ansible Galaxy idea :+1:

Issue #9, opened 2015-08-19
External role for Mongo


armab:
Use external 3rd party Ansible Galaxy role for Mongo.
Mark it as dependency for `st2` role.

armab:
Makes no sense,
since there is no good enough role for Mongo, supporting all 4 platforms we need.

PR #19, opened 2015-08-23
Replace local MongoDB role with external Ansible Galaxy dependency

armab:
Implements #9

Use **`Stouts.mongodb`** https://galaxy.ansible.com/list#/roles/982 as a dependency.

It installs MongoDB via recommended way: http://docs.mongodb.org/manual/tutorial/install-mongodb-on-ubuntu/ and also has basic TravisCI tests.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,6 +28,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.provision :hostmanager
   end
 
+  # Ensure role dependencies are in place
+  if ['up', 'provision'].include?(ARGV.first) && !(File.directory?('roles/Stouts.mongodb'))
+    system('ansible-galaxy install -r roles/st2/requirements.yml')
+  end
+
   # Use rbconfig to determine if we're on a windows host or not.
   require 'rbconfig'
   is_windows = (RbConfig::CONFIG['host_os'] =~ /mswin|mingw|cygwin/)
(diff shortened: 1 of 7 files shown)


Vagrantfile
```diff
@@ -28,6 +28,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.provision :hostmanager
   end
 
+  # Ensure role dependencies are in place
+  if ['up', 'provision'].include?(ARGV.first) && !(File.directory?('roles/Stouts.mongodb'))
```
jfryman: Is the check for role dependencies going to grow? This statement could get a bit gnarly, or there could be many of them copy/pasted. 

In what other ways might this be accomplished?

jfryman:
Nice. Big fan of upstream consumption.

Vagrantfile
armab: Yeah, don't like it either. For now it's simple workaround to see how 3rd party `mongo` role works.

* One of the solutions is using 3rd party Vagrant plugin [vagrant-host-shell plugin](https://github.com/phinze/vagrant-host-shell)
As a result installing Ansible Galaxy dependencies will be part of provision, which is more clear (`File.directory` will be removed too).
* Another solution is to create another Ansible playbook to install dependencies.

It's pretty strange that Ansible can't download dependencies automatically.
BTW, good point, thanks for taking a look :+1:

armab:
Me too, sadly 3rd party Ansible roles are not so good as Puppet or Chef recipes.

Even the most popular roles usually support no more than 1-2 Linux distros.

armab:
Closing, since external Galaxy roles dont't support all platforms we need.

PR #29, opened 2015-10-08
Ensure that the /var/log/mistral dir exists

johandahlberg:
Mistral wouldn't start on a clean vagrant box because the log dir didn't exist. This should solve that issue.


jfryman:
Excellent. Thanks for the contribution!

PR #31, opened 2015-10-09
Fixed problems in installing mistral

johandahlberg:
I added a prefix to the mistral version, and I also added a `touch /etc/mistral/database_setup.lock` to make sure that the database is not populated twice.

### DIFF
--- a/roles/mistral/tasks/sync.yml
+++ b/roles/mistral/tasks/sync.yml
@@ -3,3 +3,4 @@
   args:
     chdir: /opt/openstack/mistral
     creates: /etc/mistral/database_setup.lock
+  sudo: yes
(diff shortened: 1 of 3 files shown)


johandahlberg:
I missed #30, so the first commit isn't really anything new. However I don't know what you think of 784061b as a solution to the [duplicate db problem](https://github.com/StackStorm/st2/issues/2062#issuecomment-146847237).

jfryman:
Interesting error in CI though...

> stderr: touch: cannot touch `/etc/mistral/database_setup.lock': Permission denied

roles/mistral/tasks/sync.yml
```diff
@@ -1,5 +1,5 @@
 - name: Sync database
-  shell: /opt/openstack/mistral/.venv/bin/python ./tools/sync_db.py --config-file /etc/mistral/mistral.conf
+  shell: /opt/openstack/mistral/.venv/bin/python ./tools/sync_db.py --config-file /etc/mistral/mistral.conf && touch /etc/mistral/database_setup.lock
```
armab: Yeah, that's what I missed here https://github.com/armab/ansible-st2/commit/ce1177a17abafd0e681c57a4e7333c0e7c2f5455, good catch.

We can add `sudo: yes` for this task or go with `/opt/openstack/mistral/database_setup.lock` here and there.

armab:
Thanks!

PR #33, opened 2015-12-02
Revision can be set manually and mistral config fixes

johandahlberg:
Two small fixes:
 - If the `st2_revision` was not set to `current` `_st2_revision` was not set and thus failed downstream.
 - Made it possible to configure some of the values for mistral, i.e. the password, server and db values.

### DIFF
--- a/roles/mistral/defaults/main.yml
+++ b/roles/mistral/defaults/main.yml
@@ -1,1 +1,4 @@
 mistral_version: 0.13
+mistral_db_password: StackStorm
+mistral_db_server: localhost
+mistral_db: mistral
--- a/roles/mistral/defaults/main.yml
+++ b/roles/mistral/defaults/main.yml
@@ -1,5 +1,4 @@
 mistral_version: 0.13
-mistral_username: mistral
+mistral_db_username: mistral
 mistral_db_password: StackStorm
-mistral_db_server: localhost
 mistral_db: mistral
--- a/roles/mistral/defaults/main.yml
+++ b/roles/mistral/defaults/main.yml
@@ -1,4 +1,5 @@
 mistral_version: 0.13
+mistral_username: mistral
 mistral_db_password: StackStorm
 mistral_db_server: localhost
 mistral_db: mistral
--- a/roles/mistral/tasks/config.yml
+++ b/roles/mistral/tasks/config.yml
@@ -1,16 +1,10 @@
-- name: Copy Mistral config
+- name: Add Mistral config
   sudo: true
-  copy:
-    src: mistral.conf
+  template:
+    src: mistral.conf.j2
     dest: /etc/mistral/mistral.conf
-
-- name: Set mistral config values
-  sudo: true
-  ini_file:
-    dest=/etc/mistral/mistral.conf
-    section=database
-    option=connection
-    value=postgresql://mistral:{{mistral_db_password}}@{{mistral_db_server}}/{{mistral_db}}
+  notify:
+    - restart mistral
 
 - name: Copy Mistral log config
   sudo: true
--- a/roles/mistral/tasks/config.yml
+++ b/roles/mistral/tasks/config.yml
@@ -4,6 +4,14 @@
     src: mistral.conf
     dest: /etc/mistral/mistral.conf
 
+- name: Set mistral config values
+  sudo: true
+  ini_file:
+    dest=/etc/mistral/mistral.conf
+    section=database
+    option=connection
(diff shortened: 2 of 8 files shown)


roles/st2/tasks/2.version.yml
```diff
@@ -32,6 +32,12 @@
   when: st2_revision == 'current'
   tags: [st2, version]
 
+- name: version | Set which StackStorm revision to install manually
+  set_fact:
+    _st2_revision: {{ st2_revision }}
```
armab: From CircleCI build https://circleci.com/gh/StackStorm/ansible-st2/24:

```
ERROR: Syntax Error while loading YAML script, /home/ubuntu/ansible-st2/roles/st2/tasks/2.version.yml
Note: The error may actually appear before this position: line 37, column 21

  set_fact:
    _st2_revision: {{ st2_revision }}
                    ^
We could be wrong, but this one looks like it might be an issue with
missing quotes.  Always quote template expression brackets when they
start a value. For instance:

    with_items:
      - {{ foo }}

Should be written as:

    with_items:
      - "{{ foo }}"
```

Pretty awesome error handling from Ansible here :)

So we need just:
```yaml
_st2_revision: "{{ st2_revision }}"
```

roles/mistral/tasks/config.yml
```diff
@@ -4,6 +4,14 @@
     src: mistral.conf
     dest: /etc/mistral/mistral.conf
 
+- name: Set mistral config values
+  sudo: true
+  ini_file:
+    dest=/etc/mistral/mistral.conf
```
armab: It's better to handle/template variables in this existing file `/etc/mistral/mistral.conf`, see `st2/templates` for example.

roles/st2/tasks/2.version.yml
```diff
@@ -32,6 +32,12 @@
   when: st2_revision == 'current'
```
armab: This line and entire block below `name: version | Set which StackStorm revision to install manually` is uneeded.

Instead of that, we'll need to remove this line to pickup manual revision. After that both will work:
https://downloads.stackstorm.net/releases/st2/1.2dev/debs/current/VERSION.txt
`74`

https://downloads.stackstorm.net/releases/st2/1.2dev/debs/74/VERSION.txt
`74`

As a bonus by using this approach, we'll get validation if manual revision exists by fetching that url.

roles/mistral/defaults/main.yml
```diff
@@ -1 +1,4 @@
 mistral_version: 0.13
+mistral_db_password: StackStorm
```
armab: This is not enough, we need also to create database with these credentials in PostgreSQL.
See: https://github.com/StackStorm/ansible-st2/blob/bde33777da7b2327bc0571418bf720ce43f7f716/roles/mistral/meta/main.yml#L20-L24

roles/mistral/defaults/main.yml
```diff
@@ -1 +1,4 @@
...
+mistral_db_server: localhost
```
armab: Are you using external db (not `localhost`)?

Because `ANXS.postgresql` dependency https://github.com/StackStorm/ansible-st2/blob/bde33777da7b2327bc0571418bf720ce43f7f716/roles/mistral/meta/main.yml#L15-L18 will install/configure PostgreSQL on localhost.

johandahlberg:
@armab Thanks for all the comments. I was a bit quick to push send on this and didn't fully think things through. I'll go through it tomorrow and update the PR.

armab:
Thank you!
Anyway, overall good changes in your PR :+1:

johandahlberg:
I've fixed this according to your comments now.

I've also added another small feature. `st2_system_user_in_sudoers` can now be set to `no` if you don't want the `st2_system_user` to be in sudoers, which we don't want on our production systems.

Let me know if you want me to rebase this and clean up the commit log right now it's a bit of a mess reflecting my process to get everything working.

roles/st2/tasks/3.user.yml
```diff
@@ -29,6 +29,6 @@
     mode: 0440
     regexp: "^{{ st2_system_user }} ALL="
     line: "{{ st2_system_user }} ALL=(ALL) NOPASSWD: SETENV: ALL"
-    state: present
+    state: "{{ 'present' if st2_system_user_in_sudoers else 'absent' }}"
```
armab: Nice! :+1:

armab:
Squashing/rebasing is not required, everything is ok :+1:
@johandahlberg Nice coding decisions, thank you for PR and your time!

I'll add new config settings in docs.

PR #35, opened 2015-12-07
Version the pip dependencies

johandahlberg:


roles/st2/tasks/4.dependencies.yml
```diff
@@ -27,5 +27,5 @@
 - name: dependencies | Install st2 pip dependencies
   sudo: true
   pip:
-    requirements: https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt
+    requirements: "https://raw.githubusercontent.com/StackStorm/st2/v{{ st2_version }}/requirements.txt"
```
armab: @johandahlberg Another good fix!

I got your idea, use: 
`https://downloads.stackstorm.net/releases/st2/{{ _st2_version }}/requirements.txt`

See also for better understanding: 
https://github.com/johandahlberg/ansible-st2/blob/make_dep_versioned/roles/st2/tasks/2.version.yml#L24

armab:
Thanks @johandahlberg,
good catch!

PR #37, opened 2016-01-12
Don't delegate version lookup to localhost

### LINKED ISSUE #36: Literal version lookup fails
The "Lookup literal version" task is failing silently with "sort: unrecognized option `--version-sort'"
Leaving _st2_version set to "" and breaking following tasks

rjmackay:
Delegating version lookup to localhost breaks when the localhost doesn't have standard GNU coreutils (ie. on OSX)
Just run it on the remote host instead.

Fixes #36

### DIFF
--- a/roles/st2/tasks/2.version.yml
+++ b/roles/st2/tasks/2.version.yml
@@ -11,7 +11,6 @@
     uniq | head -n 1
   when: st2_version == 'stable' or st2_version == 'unstable'
   register: _st2_version
-  delegate_to: 127.0.0.1
   changed_when: False
   tags: [st2, version]
 


armab:
There was a good intent to do this lookup locally, but if it fixes your issue - let's change it :+1: 

There is another `delegate_to: 127.0.0.1` in the same file you edited.
Let's remove it too to be consistent, and we're ready to Merge.

----
Build failed, but not realted to your change
(I'll need to verify all playbooks later to be compatible with most recent Ansible version).

armab:
Not needed anymore since installation story migrated to new #42 packages: https://stackstorm.com/2016/05/03/road-to-packages/

PR #42, opened 2016-04-01
Updating to use deb packages

johandahlberg:
I've update the roles and playbook to use the new deb-packages to install st2. Essentially I've converted the [install guide](https://docs.stackstorm.com/install/deb.html) to a playbook.

Sorry for the massive changes. I've tried as much as possible to dived this into commits that make sense, but I might have mixed things up in some places. I can now build this in the supplied Vagrant box, and I've also added a small smoke test role, which should confirm that some basic things are working as expected. 

There are probably some things which still need fixing, especially in relation to making sure that all customization relating to the variables are working, but this is at least a working draft.

PR #44, opened 2016-05-06
Docs cleanup for #42

armab:
Synchronize docs with recent #42 which installs new packages


README.md
```diff
@@ -41,32 +40,28 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `mistral_db_password` | `StackStorm`  | PostgreSQL DB password for Mistral.
```
armab: Looks like big room for improvement for number of variables/settings.
For now OK as something minimum viable.

PR #46, opened 2016-05-11
Use correct APT syntax for specifying version revisions

johanherman:
Minor fix to use the correct syntax when specifying version revisions to apt.


armab:
Good catch, thanks! :+1:

Issue #49, opened 2016-08-23
Switch to MongoDB 3.2


armab:
Since StackStorm `v1.6` MongoDB `3.2` is the default version,
see: https://docs.stackstorm.com/install/deb.html#install-dependencies for installation instructions.

Reflect the change in st2 Configuration Management repos.

PR #51, opened 2016-08-26
Switch to MongoDB 3.2 packages

armab:
> Work on #49 

Since [StackStorm v1.6](https://stackstorm.com/2016/08/10/stackstorm-v1-6-0/), MongoDB `3.2` is the default supported version.


lakshmi-kannan:
👍

PR #53, opened 2016-10-14
Fix Ansible 2.1 become-unprivileged-user compatibility issue

unicell:
Enable per task pipelining to fix issue with Ansible 2.1 [1], otherwise it
would faile with "Failed to set permissions on the temporary files
Ansible needs to create when becoming an unprivileged user."

[1] https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user"

### DIFF
--- a/roles/mistral/tasks/main.yml
+++ b/roles/mistral/tasks/main.yml
@@ -32,6 +32,8 @@
     - restart mistral
 
 - name: Initiate database
+  vars:
+      ansible_ssh_pipelining: true
   become: yes
   become_user: postgres
   shell: psql < /etc/mistral/init_mistral_db.SQL
--- a/roles/st2/tasks/user.yml
+++ b/roles/st2/tasks/user.yml
@@ -13,6 +13,8 @@
   tags: [st2, user]
 
 - name: user | Authorize key-based access for system user
+  vars:
+      ansible_ssh_pipelining: true
   sudo: yes
   sudo_user: "{{ st2_system_user }}"
   authorized_key:


armab:
Thanks!
I'll dig into this and review later.

Looking really quick, our CI couldn't catch it, see: 
- https://travis-ci.org/StackStorm/ansible-st2/builds/156585571
- Ubuntu14: https://travis-ci.org/StackStorm/ansible-st2/jobs/156585572#L1085
- Ubuntu16: https://travis-ci.org/StackStorm/ansible-st2/jobs/156585573#L1128

Apart of that, some questions to help me clarifying it.
How can I reproduce it? What are prerequisites?
`Ansible 2.1`, something else? Will this bug show in current [`Vagrant`](https://github.com/StackStorm/ansible-st2/blob/master/Vagrantfile) environment?

@unicell I suspect this happens when running Ansible non-locally, eg. for remote machines, is this your case? 
That could be the reason why It wasn't caught by CI, will dig later how current ansible + [`test-kitchen`](https://github.com/StackStorm/ansible-st2/blob/master/.kitchen.yml) works.

unicell:
@armab According to the Note section of https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user , the issue only happens when it is unprivileged user for BOTH the one who made the connection AND the one it is trying to switch to.

The latter one is always true in the playbook (`postgres` and `"{{ st2_system_user }}"`). I was using my own user (non root) when running the playbook, that triggers the problem. Not sure whether remote or local matters, but I do run the playbooks for remote boxes.

FYI, I'm also hitting the same issue for the upstream `ANXS.postgresql` module when apply the playbook. Filed an issue for them as well, https://github.com/ANXS/postgresql/pull/211 But since they're enforcing `--syntax-check` in CI and override per-task vars for older version of Ansible will fail this syntax check. Haven't figure out way to properly address.

PR #54, opened 2016-12-04
Make 'kitchen-docker' work in OSX.

jvrplmlmn:
Reference:
- https://github.com/test-kitchen/kitchen-docker/issues/207
- https://github.com/test-kitchen/kitchen-docker/issues/201

Solves:

image

WDYT @armab ?

PR #55, opened 2016-12-04
Fix PostgreSQL build failure

armab:
Discovered in https://github.com/StackStorm/ansible-st2/pull/54

```sh
TASK [ANXS.postgresql : PostgreSQL | Make sure the PostgreSQL users are present] ***

       task path: /tmp/kitchen/roles/ANXS.postgresql/tasks/users.yml:8

       fatal: [localhost]: FAILED! => {"failed": true, "msg": "the field 'args' has an invalid value, which appears to include a variable that is undefined. The error was: 'unicode object' has no attribute 'name'\n\nThe error appears to have been in '/tmp/kitchen/roles/ANXS.postgresql/tasks/users.yml': line 8, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: PostgreSQL | Make sure the PostgreSQL users are present\n  ^ here\n"}
```

This is probably due to newer Ansible and older dependency.
Try to Update external dependency `ANXS.postgresql` to latest.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -15,7 +15,8 @@ provisioner:
     # For the moment only staging repos have Ubuntu Xenial st2 packages
     # TODO: Remove when Ubuntu Xenial packages are available in prod
     st2_pkg_repo: staging-stable
-  ansible_version: latest
+  require_pip: true
+  ansible_version: 2.1.2
   ansible_verbose: true
   ansible_verbosity: 2
   # TODO: Make sure Ansible playbooks are idepotent
--- a/playbooks/requirements.yml
+++ b/playbooks/requirements.yml
@@ -1,6 +1,10 @@
 ---
 - hosts: all
   tasks:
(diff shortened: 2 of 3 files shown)


armab:
So it's Ansible `2.2` incompatibility with external role `ANXS.postgresql`: https://github.com/ANXS/postgresql/issues/218

Pinning Ansible to `2.1` for now.

jvrplmlmn:
@armab Seems like there is a discrepancy between the `ANSX.postgresql` version in`roles/mistral/meta/main.yml` and `roles/mistral/requirements.yml`. I'm not very familiar with `ansible-galaxy` and the ansible dependency system, but I think that Kitchen is using `v1.2.1` instead of `v1.6.2`.

armab:
@jvrplmlmn Good catch!

Opened https://github.com/StackStorm/ansible-st2/pull/56 for that

PR #56, opened 2016-12-05
Update ANXS.postgresql for real

armab:
See: https://github.com/StackStorm/ansible-st2/pull/55

Thanks @jvrplmlmn for pointing to that!

It looks even more messed. Latest ANXS.postgresql on GitHub is `v1.6.2`, while latest released on Galaxy is `v1.5.0`.

Pinning to version on galaxy.


jvrplmlmn:
>        - downloading role from https://github.com/ANXS/postgresql/archive/v1.5.0.tar.gz

👍

Issue #15, opened 2015-08-19
More Rspec Tests


armab:
Before publishing on Ansible Galaxy, add more Rspec Integration tests, increase infrastructure coverage.

Continue work started at https://github.com/StackStorm/ansible-st2/issues/7

armab:
Some facts from #22 #7 research, not to forget about: 

* CircleCI supports Docker
* [Travis now supports Docker too](http://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/)
* KitchenCI:
  * [kitchen-docker](https://github.com/portertech/kitchen-docker)
  * [kitchen-vagrant](https://github.com/test-kitchen/kitchen-vagrant)
  * [kitchen-ansible](https://github.com/neillturner/kitchen-ansible)
* But docker way is not good enough for current environment with many services installed
* (!) It's possible to run `lxc` containers under CircleCI
* [It's not possible to run `lxc` containers under Travis](https://travis-ci.org/armab/travis-lxc/builds)
* (!) [New Travis infra can run Virtualbox + Vagrant VMs](https://travis-ci.org/armab/travis-vagrant/builds/77223187), but only for x32 images (sad)
* To dig further:
  * [Vagrant-lxc as KitchenCI driver under CircleCI](https://github.com/fgrehm/vagrant-lxc/issues/339)
  * [kitchen-lxc](https://github.com/chrisroberts/kitchen-lxc) under CircleCI
  * play with `lxc` containers under CircleCI ssh, try to build custom `lxc` templates

As a last solution: use [kitchen-digitalocean](https://github.com/test-kitchen/kitchen-digitalocean) or [kitchen-ec2](https://github.com/test-kitchen/kitchen-ec2) to spawn up several machines for tests. Not free, but still very cheap, especially considering pay-per-hour model.

armab:
If running CitchenCI + AWS (to test against different distros) - consider using AWS Spot Instances with new `BlockDuration` option, which guarantees running low-priced spot instance for some minimum `--block-duration-minutes`.
https://aws.amazon.com/blogs/aws/new-ec2-spot-blocks-for-defined-duration-workloads/

This is cheap & perfect for testing, when working instance is needed for ~15 minutes or so.

armab:
Some things to consider about testing Ansible roles in Docker: 
https://www.ansible.com/blog/testing-ansible-roles-with-docker
- use Docker images with init/upstart (same as we do in `st2-packages` repo)
[...]

PR #57, opened 2016-12-05
Add InSpec as a verifier for KitchenCI + testing for the role st2/user

jvrplmlmn:
Brings https://github.com/StackStorm/chef-stackstorm/issues/47 to this repository.

This is part of https://github.com/StackStorm/ansible-st2/issues/15

WDYT @armab ?

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -7,6 +7,9 @@ driver:
 transport:
   name: sftp
 
+verifier:
+  name: inspec
+
 provisioner:
   name: ansible_playbook
   hosts: all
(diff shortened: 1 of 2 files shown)


armab:
Thanks @jvrplmlmn But not yet for this repo.

Also, there is no guarantee it will be InSpec for `ansible-st2` (due to internal discussions/plans).

Even though, it might be useful in future, so let's keep this PR opened :+1:

jvrplmlmn:
That's totally fine. I was playing with using InSpec in Ansible, so I though why not pushing this here. Feel free to close this PR if it is not needed in the future.

PR #61, opened 2016-12-21
Minor fixes

Mierdin:
- Fixed issue with roles location (pwd would be `ansible-st2/playbooks` so needs to go up a directory first to get to `ansible-st2/roles`)
- Added .retry files to .gitignore


armab:
Looks good :+1: 

Also FYI: https://github.com/StackStorm/ansible-st2/blob/master/ansible.cfg#L2

Mierdin:
@armab I noted your reference to the ansible.cfg file, thanks for that. I suspect it's not getting picked up when I run with Vagrant, because before my changes here, I was getting this error:

```
==> default: Running provisioner: ansible...
    default: Running ansible-playbook...

PLAY [all] *********************************************************************

TASK [setup] *******************************************************************
ok: [default]

TASK [Install galaxy dependencies] *********************************************
changed: [default]

TASK [show message] ************************************************************
ok: [default] => {
    "msg": {
        "changed": true,
        "cmd": [
            "ansible-galaxy",
            "install",
            "-r",
            "roles/mistral/requirements.yml"
        ],
        "delta": "0:00:00.375551",
        "end": "2016-12-21 15:01:49.428682",
        "rc": 0,
        "start": "2016-12-21 15:01:49.053131",
        "stderr": "\u001b[0;31m [ERROR]: Unable to open roles/mistral/requirements.yml: [Errno 2] No such file\nor directory: u'roles/mistral/requirements.yml'\n\u001b[0m",
        "stdout": "",
        "stdout_lines": [],
        "warnings": []
    }
}

PLAY RECAP *********************************************************************
default                    : ok=3    changed=1    unreachable=0    failed=0

==> default: Running provisioner: ansible...
```

Or, in case it was being picked up but the path was wrong, I tried this (still didn't work):

```
[defaults]
roles_path=../roles/
```

Mierdin:
@armab Ignore my last - I had my streams crossed. The problem is that the pwd will always be `playbooks/` and since setup.cfg is a directory above that, it won't be seen unless explicitly referenced. My suggestion is to move this configuration file into the `playbooks/` directory.

Mierdin:
So here's the long and short of it: it looks like the working directory will always be `playbooks/`, and there's no `ansible.cfg` there. So, the playbooks need to be in the same directory as `ansible.cfg`.

My current approach was to simply move everything out of `playbooks/`. [Seems to work right now](https://gist.github.com/Mierdin/ad944b63e05af505c1730216100a0733)

armab:
@Mierdin That's really strange, because default `Vagrantfile` we have in place worked for me. Just re-ran it again to be sure, - no problems.

What's your `Vagrant` version and `Ansible` version BTW?
```sh
ansible 2.2.0.0
Vagrant 1.8.6
```
It could be rather my old Vagrant version or your old version. Probably something changed between the versions, like Ansible provisioner implementation in Vagrant. Curious about any real reasons, because the fact that you can't run playbook from the sub-directory could be a sign of some deeper inconsistencies. Additionally, there is CI which re-uses same `stackstorm.yml` from non-root dir.

The reason I place example playbooks in sub-directory is that they're provided just for reference and example, which I would like to preserve if possible.
It's OK for users to create their own playbook where they can list [roles](https://github.com/StackStorm/ansible-st2/tree/master/roles) from this repo, as well as provide special parameters for these roles.

Anyways, if we won't find anything interesting and that's the only way to fix things for you, - I'm OK with that as well.

Vagrantfile
```diff
@@ -28,9 +28,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   config.vm.provision :ansible do |ansible|
-    ansible.playbook = "playbooks/requirements.yml"
+    ansible.playbook = "requirements.yml"
   end
   config.vm.provision :ansible do |ansible|
-    ansible.playbook = "playbooks/stackstorm.yml"
+    ansible.playbook = "stackstorm.yml"
```
armab: Please update `README.md` with the new path.

bigmstone:
@armab Wouldn't work for me either. Running ansible 2.2.0 and vagrant 1.9.0

armab:
@bigmstone Thanks! This might be related to Vagrant version (`1.8.6` vs `1.9.0`). @Mierdin Please confirm you have `1.9.x` as well.

So if it's Vagrant version logic, I'm OK with change, just update `README.md` and LGTM.

Mierdin:
Hrm, I have `Vagrant 1.8.6`

PR #62, opened 2016-12-22
Moved to Xenial, added step to install ansible when running in vagrant

Mierdin:


armab:
Why we're moving to Xenial Vagrant in `ansible-st2`?
I don't think it's "default" OS to install StackStorm and not default OS for Ansible playbooks development as well due to different Vagrant-related problems with new Xenial, when random things might not work for different users/environments.

I follow the same logic in `chef-stackstorm` repository and would like to keep consistency in Configuration Management repos I maintain for a long time. The recommended OS was always `Ubuntu14`.

What will work for me if you could provide more extended `Vagrantfile` logic for both OSes: 
```sh
vagrant up ubuntu14
vagrant up ubuntu16

# but keep ubuntu14 default
vagrant up
```

@lakstorm @Mierdin Please care to collaborate before merging.
Additionally, don't forget to include decent description in every PR.

PR #63, opened 2017-01-04
Add .vagrant and ANXS.postgresql to .gitignore

humblearner:


Vagrantfile
```diff
@@ -7,6 +7,7 @@ VAGRANTFILE_API_VERSION = '2'
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = 'ubuntu/xenial64'
   config.vm.hostname = "#{HOSTNAME}"
+  config.vm.network "forwarded_port", guest: 22, host: rand(30000)+1234
```
humblearner: In case vagrant port 2222 is occupied by some other vagrant instance

armab: :-1: for doing `random` hack like that.
As I remember, if `2222` is used, Vagrant will try to take `2200` instead and so on.

@humblearner what `Vagrant` version do you have?

.gitignore
```diff
@@ -1,3 +1,10 @@
...
-*.retry
\ No newline at end of file
+*.retry
+
+# Ansible Galaxy Role
+roles/ANXS.postgresql/**
+
+# Vagrant
+.vagrant/
+ubuntu-xenial-16.04-cloudimg-console.log
```
armab: Maybe it worth masking all `*.log` files here for future?

Vagrantfile
armab: :-1: for doing random hack like that.

As I remember if `2222` is used, Vagrant will try to use `2200` instead and so on.

@humblearner what `Vagrant` version do you have?

humblearner:
```
$ vagrant --version
Vagrant 1.8.6
```
Sure, I will try the official way: `auto_correct: true` from: https://www.vagrantup.com/docs/networking/forwarded_ports.html

Vagrantfile
```diff
@@ -7,6 +7,7 @@ VAGRANTFILE_API_VERSION = '2'
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = 'ubuntu/xenial64'
   config.vm.hostname = "#{HOSTNAME}"
+  config.vm.network "forwarded_port", guest: 22, host: 2200, auto_correct: true
```
humblearner:
```
==> centos72: Fixed port collision for 22 => 2200. Now on port 2203.
==> centos72: Fixed port collision for 22 => 2222. Now on port 2204.
==> centos72: Clearing any previously set network interfaces...
==> centos72: Preparing network interfaces based on configuration...
    centos72: Adapter 1: nat
    centos72: Adapter 2: hostonly
==> centos72: Forwarding ports...
    centos72: 22 (guest) => 2203 (host) (adapter 1)
    centos72: 22 (guest) => 2204 (host) (adapter 1)
==> centos72: Running 'pre-boot' VM customizations...
==> centos72: Booting VM...
==> centos72: Waiting for machine to boot. This may take a few minutes...
    centos72: SSH address: 127.0.0.1:2204
```

armab: I was sure Vagrant tries to fix any SSH port collision by default.

If that wouldn't happen, then we'd change all Vagranfiles we have in all repos:
https://github.com/search?q=org%3AStackStorm+Vagrantfile&type=Code including https://github.com/StackStorm/st2vagrant/blob/master/Vagrantfile

So for me it seems like we have some unexpected Vagrant usage here.
How do you run this Vagrant VM?
What's your Vagrant version?

This also might help to understand the problem better:
http://stackoverflow.com/a/24131225/4533625
http://stackoverflow.com/a/25142933/4533625

humblearner: @armab: You are right, we need this change in all the `VagrantFile`s (Consider this as a starting point. ). Generally I face this problem, when i have multiple instances of vagrant running from different repos. For example: Any other vagrant instance fails, if I have vagrant instance from `ansible-st2` up, with following error: 
```
$ vagrant up centos72
Bringing machine 'centos72' up with 'virtualbox' provider...
==> centos72: Checking if box 'bento/centos-7.2' is up to date...
==> centos72: A newer version of the box 'bento/centos-7.2' is available! You currently
==> centos72: have version '2.3.0'. The latest is version '2.3.1'. Run
==> centos72: `vagrant box update` to update.
Vagrant cannot forward the specified ports on this VM, since they
would collide with some other application that is already listening
on these ports. The forwarded port to 2200 is already in use
on the host machine.

To fix this, modify your current project's Vagrantfile to use another
port. Example, where '1234' would be replaced by a unique host port:

  config.vm.network :forwarded_port, guest: 22, host: 1234

Sometimes, Vagrant will attempt to auto-correct this for you. In this
case, Vagrant was unable to. This is usually because the guest machine
is in a state which doesn't allow modifying port forwarding. You could
try 'vagrant reload' (equivalent of running a halt followed by an up)
so vagrant can attempt to auto-correct this upon booting. Be warned
that any unsaved work might be lost. 
```
I am not in favor of `vagrant reload`. Because of RAM limitation, I can have a maximum of 2 instances running. Have to suspend them in case i want to bring up third instance. Repos with Vagrant: `st2vagrant`, `ansible-st2`, `st2` dev env., `st2-test-ground`. Last two have multiple vagrants (based on os flavor). If you still think this is not right, i can revert this line. :)

armab: Thanks for detailed explanation!
I'm OK with the change, since the port is not random anymore.

PR #66, opened 2017-01-04
EL6/7 and CentOS 6/7 support for multiple roles

Mierdin:
This patch adds support for the following operating systems:

- CentOS 6
- CentOS 7
- RHEL 6
- RHEL 7

> I have also verified that an Ubuntu Trusty install still works after these changes.

The following roles have been created:
- `epel`

The following roles have been modified (either to support `epel` installation, or to work on EL/CentOS in general):
- `st2`
- `st2repos`
- `mongodb`
- `st2mistral`


roles/st2/tasks/main.yml
```diff
@@ -1,22 +1,10 @@
...
-- name: Install latest st2 package
-  become: yes
-  apt:
-    name: st2
-    state: latest
-  when: st2_version == "stable"
-  notify:
-    - restart st2
+- include: debian.yml
+  when: ansible_os_family == 'Debian'
```
armab: We can do 1 include without `when` condition.

See @humblearner arppoach https://github.com/StackStorm/ansible-st2/blob/3785d849135599de1f667e692576d3501af67fb5/roles/mongodb/tasks/main.yml#L3 with 2 files `mongodb_apt.yml` and `mongodb_yum.yaml` which looks nicer and cleaner.

It would be better if we follow one standard in all roles for consistency.

roles/st2/tasks/redhat.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Install latest RHEL st2 package (stable)
+  become: yes
+  yum:
+    name: st2
+    state: latest
+    disable_gpg_check: yes
```
armab: Why we have `disable_gpg_check: yes` for `stable` and nothing for `unstable` in next block? Any specific reason?

roles/st2repos/tasks/debian.yml
```diff
@@ -0,0 +1,23 @@
...
+    name: "{{ item }}"
+    state: present
+    update_cache: yes
+  with_items:
+    - debian-archive-keyring
+    - apt-transport-https
+
+- name: Add keys to keyring
+  become: yes
+  apt_key:
```
armab: I think it would be more solid if we check `id` (identifier of key) here as well.
See example with Mongo: https://github.com/StackStorm/ansible-st2/blob/3785d849135599de1f667e692576d3501af67fb5/roles/mongodb/tasks/mongodb_apt.yaml#L6

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,14 @@
...
+    state: present
+    key: https://packagecloud.io/StackStorm/stable/gpgkey
+
+- name: Add ST2 Repo
+  become: yes
+  yum_repository:
+    name: stackstorm
+    description: Stackstorm Repo
+    file: st2
+    baseurl: https://packagecloud.io/StackStorm/stable/el/$releasever/$basearch
```
armab: Missing `{{ st2_pkg_repo }}` variable here which is responsible for repository we install (`stable`/`unstable` etc).

See: https://github.com/StackStorm/ansible-st2#variables and existing rule in `deb`: https://github.com/StackStorm/ansible-st2/pull/66/files#diff-37045cfaa2d3524070e708be840837adR21

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,14 @@
...
+  rpm_key:
+    state: present
+    key: https://packagecloud.io/StackStorm/stable/gpgkey
+
+- name: Add ST2 Repo
+  become: yes
+  yum_repository:
+    name: stackstorm
+    description: Stackstorm Repo
+    file: st2
```
armab: Here is yum repo file I got for `CentOS7` when adding st2 packagecloud repo: 

```sh
[vagrant@centos7 ~]$ cat /etc/yum.repos.d/StackStorm_stable.repo
[StackStorm_stable]
name=StackStorm_stable
baseurl=https://packagecloud.io/StackStorm/stable/el/7/$basearch
repo_gpgcheck=1
gpgcheck=0
enabled=1
gpgkey=https://packagecloud.io/StackStorm/stable/gpgkey
sslverify=1
sslcacert=/etc/pki/tls/certs/ca-bundle.crt
metadata_expire=300
```

A lot of security checks here by PackageCloud. I think we should include more `yum_repository` options where it's possible and looks easy.

Additionally, it would be nice to follow the same naming for the file and repo.

roles/st2/tasks/main.yml
Mierdin: Fixed in 6a4467f

roles/st2/tasks/redhat.yml
Mierdin: No reason, just missed it. Added in 5a1f2e4

roles/st2repos/tasks/redhat.yml
Mierdin: Added in c40a809

roles/st2repos/tasks/redhat.yml
Mierdin: So would you prefer I undo https://github.com/StackStorm/ansible-st2/pull/66/commits/8f1e76b94b974379ead556b67f8fec2ec170338b? I was using the script to automatically generate this configuration previously

roles/st2repos/tasks/debian.yml
Mierdin: Where would I go about finding that?

Mierdin: Nevermind, added in f6ac495

Mierdin:
@armab I don't see anything that would prevent CentOS6 from running, I just haven't tested it. I'll test and update the title accordingly, this should allow both.

Also I am on board for modifying CI before merging this.

roles/st2/tasks/config_auth.yml
```diff
@@ -1,11 +1,29 @@
...
     state: present
   with_items:
     - python-passlib
     - apache2-utils
+  when: ansible_os_family == 'Debian'
+
+- name: Install epel-release repo (RedHat)
+  become: yes
+  yum:
+    name: epel-release
```
armab: Will this work on real `RHEL6`?
You can test in AWS https://aws.amazon.com/marketplace/pp/B00CFQWLS6

roles/st2/tasks/main.yml
```diff
@@ -1,22 +1,6 @@
...
-
-- name: Install latest st2 package
-  become: yes
-  apt:
-    name: st2={{ st2_version }}-{{ st2_revision }}
-    state: present
-  when: st2_version != "stable"
-  notify:
-    - restart st2
+- include: "{{ ansible_os_family }}.yml"
```
armab: Since it only about installing package, you can avoid splitting into `{{ ansible_os_family }}.yml` here, just use `package` module. http://docs.ansible.com/ansible/package_module.html

See solution in `st2mistral` by @humblearner https://github.com/StackStorm/ansible-st2/blob/1426c6d28184dc61f6d41cb7b2731966d331d3bf/roles/st2mistral/tasks/main.yml#L10-L24 which is pretty elegant approach.

roles/st2repos/tasks/debian.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Install prereqs (Debian)
+  become: yes
+  apt:
+    name: "{{ item }}"
+    state: present
+    update_cache: yes
```
armab: This will break Ansible Idempotence as @bigmstone discovered before, since it will `apt-get upgrade` every time.

roles/st2repos/tasks/redhat.yml
armab: I'm for hardcoding safe values we can use easily via `yum_repository`, without auto-generating them.

roles/st2/tasks/config_auth.yml
Mierdin: Yes, I will include both EL6 and EL7 in my testing.

roles/st2/tasks/main.yml
Mierdin: Done in https://github.com/StackStorm/ansible-st2/pull/66/commits/3c1fd1436946371d73a8ada257492b21d7e54995

roles/st2repos/tasks/debian.yml
Mierdin: Removed in 025ed66

README.md
```diff
@@ -64,6 +64,20 @@ You might be interested in other methods to deploy StackStorm engine:
   * [RHEL7/CentOS7](https://docs.stackstorm.com/install/rhel7.html)
   * [RHEL6/CentOS6](https://docs.stackstorm.com/install/rhel6.html)
 
+## Developing
```
armab: :+1:

meta/main.yml
```diff
@@ -14,6 +14,14 @@ galaxy_info:
       versions:
         - trusty
         - xenial
+    - name: EL
```
armab: Don't change yet.

We'll update it later on final stage, after this PR, `st2web` PR and after disabling `allow_failures` for CentOS6/7 CI: https://github.com/StackStorm/ansible-st2/blob/master/.travis.yml#L14

roles/epel/meta/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  platforms:
+    - name: Ubuntu
+      versions:
+        - trusty
+        - precise
+    - name: EL
+      versions:
+        - 6
+        - 7
+    - name: CentOS
```
armab: Not needed.
Previous name: `EL` covers both RHEL and CentOS.

At least I couldn't find occurencies like that in any galaxy role.

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,21 @@
...
+  set_fact:
+      rh_ver: "{{ ansible_distribution_version|int|abs }}"
+  # TODO(mierdin): Remove when https://github.com/StackStorm/ansible-st2/issues/89 is addressed
+
+- name: Add ST2 Repo
+  become: yes
+  yum_repository:
+    name: "StackStorm_{{ st2_pkg_repo }}"
+    description: "StackStorm_{{ st2_pkg_repo }}"
+    file: st2
```
armab: Naming should follow the PackageCloud script install logic where possible.
I see `/etc/yum.repos.d/StackStorm_stable.repo` for my installation.

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Define RH Version
+  set_fact:
+      rh_ver: "{{ ansible_distribution_version|int|abs }}"
```
armab: Is converting to `int` and so on required?
`ansible_distribution_version` shows `7Server` for you?

Mierdin: It was before I discovered `ansible_distribution_major_version`. This was my way of converting `7.3` to `7`.

`7Server` is something that comes up on the server side, not an Ansible thing. So actually, I'm using `ansible_distribution_version`, which equals `7.3`, and converting it to `7` with these filters in order to make sure the right major version was used in the yum config (instead of relying on the yum variable `releasever`, which did result in `7Server`).

Anyways, I just used `ansible_distribution_major_version` instead of these filters in 6984fae. Achieves same result without the filters

armab: :+1:

roles/epel/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+- name: Download EPEL RPM
+  get_url:
+    url: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm"
+    dest: "/tmp/epel{{ ansible_distribution_major_version }}.rpm"
+    mode: 0440
+  when: not st.stat.exists and ansible_distribution == "RedHat"
+
+- name: Install EPEL repo
+  yum:
+    name: "/tmp/epel{{ ansible_distribution_major_version }}.rpm"
```
armab: Temp file is not needed, `yum` can download URL directly, see: http://docs.ansible.com/ansible/yum_module.html

> You can also pass a url or a local path to a rpm file

roles/epel/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Check if EPEL is installed
+  stat:
+    path: /etc/yum.repos.d/epel.repo
+  register: st
```
armab: `st` is too generic and doesn't really describes well what is var about. How about `epel_is_installed` or similar?

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,16 @@
+---
+- name: Add ST2 Repo
+  become: yes
+  yum_repository:
+    name: "StackStorm_{{ st2_pkg_repo }}"
+    description: "StackStorm_{{ st2_pkg_repo }}"
+    file: st2
+    baseurl: https://packagecloud.io/StackStorm/{{ st2_pkg_repo }}/el/{{ ansible_distribution_major_version }}/$basearch
```
armab: So seems this is good enough and doesn't look as hack anymore :+1: 

I think we can close the https://github.com/StackStorm/ansible-st2/issues/89 now, just add a note that we should always use `{{ ansible_distribution_major_version }}` instead of `$releasever` when adding _any_ yum repo.

meta/main.yml
Mierdin: Removed in 15ba22e

roles/epel/meta/main.yml
Mierdin: Ack that - updated in 80032d9

roles/epel/tasks/main.yml
Mierdin: Ack that, updated in 09df69b

roles/epel/tasks/main.yml
Mierdin: Updated in 1904363

roles/st2repos/tasks/redhat.yml
Mierdin: Yup, cool. Closed #89

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,16 @@
...
+    # TODO(mierdin): Should consider moving back to $releasever once https://github.com/StackStorm/ansible-st2/issues/89 is addressed
```
armab: `TODO` line can be removed as well, since you fixed https://github.com/StackStorm/ansible-st2/issues/89

roles/st2repos/tasks/redhat.yml
Mierdin: Fixed in 7254866

roles/epel/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  stat:
+    path: /etc/yum.repos.d/epel.repo
+  register: epel_installed
+  when: ansible_distribution == "RedHat"
+
+- name: Install EPEL repo
+  yum:
+    name: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm"
+    state: present
+  when: not epel_installed.stat.exists and ansible_distribution == "RedHat"
```
armab: `ansible_distribution` could be `CentOS` so it's safer to use `ansible_os_family` here (as you did in other places).

Mierdin: Fixed in https://github.com/StackStorm/ansible-st2/pull/66/commits/da000f663a3626cb5ddd02516443e5bd04332469

roles/st2repos/tasks/redhat.yml
Mierdin: Fixed in abb465e

Mierdin:
@armab Okay to my knowledge, I've addressed all your comments, and the only EL failures going on right now are related to nginx. Should be ready to go

Mierdin:
@armab @mstone Okay for reals this time, I have tested on EL6 and EL7, and I also re-ran this on CentOS 6/7 and Ubuntu to make sure those still work.

Only outstanding issue is to make sure that I did the right thing here: https://github.com/StackStorm/ansible-st2/pull/66/commits/f5bb4b5a4dd95365f33b0587693e2220ef55e21c

I had to do this to get rid of the error: `sudo: sorry, you must have a tty to run sudo`

This was originally implemented in https://github.com/StackStorm/ansible-st2/commit/11129e980327ac9f32d108ce129abef66cc3612f but everything seems to run just fine without it.

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,22 @@
...
+    gpgcheck: no
+    enabled: yes
+    sslverify: yes
+
+# Fixes "Failure talking to yum: Cannot retrieve repository metadata (repomd.xml) for repository: StackStorm_stable. Please verify its path and try again" when installing st2
+- name: Update ca-certificates package
+  become: yes
+  package:
+    name: ca-certificates
+    state: latest
```
armab: > [ANSIBLE0010] Package installs should not use latest

The `ansible-lint` warning is valid. Use `state: installed` where possible.
Installing `latest` package version on every playbook re-run is generally a bad practice (especially when ansible works in `ansible-pull` mode).

For `st2` we use `latest` intentionally as option so user can specify `st2_version=latest` to have latest StackStorm. 
Probably we should change default behavior there someday later since it could be dangerous for default installs.

For everything else the safe way is to upgrade software on the server by running specific playbook which focuses on OS package upgrades (or patches) to have full control and react accordingly if something went wrong.

roles/st2/tasks/main.yml
```diff
@@ -1,17 +1,25 @@
 ---
-- name: Install latest st2 package
+- name: Install libffi-devel on EL6
   become: yes
-  apt:
+  package:
+    name: ftp://fr2.rpmfind.net/linux/centos/6/os/x86_64/Packages/libffi-devel-3.0.5-3.2.el6.x86_64.rpm
+    state: latest
```
armab: See comment https://github.com/StackStorm/ansible-st2/pull/66/files#r97052091

roles/st2/tasks/main.yml
```diff
@@ -1,17 +1,25 @@
...
+  when: ansible_os_family == "RedHat" and ansible_distribution_major_version == "6"
+  ignore_errors: yes  # Ignore error when already installed
```
armab: No, that's a hack.
Additionally, I'm not sure if this block is idempotent.

Please double check that we don't download `ftp://fr2.rpmfind.net/linux/centos/6/os/x86_64/Packages/libffi-devel-3.0.5-3.2.el6.x86_64.rpm` on every playbook re-run.

armab: Per https://docs.stackstorm.com/install/rhel6.html#system-requirements did you check if it's possible to enable `server-optional` repository for RHEL6?
So this dependency will be installed automatically by `st2`.

Find `libffi-devel` for `st2` package requirements: https://packagecloud.io/StackStorm/stable/packages/el/6/st2-2.1.1-1.x86_64.rpm

roles/mongodb/meta/main.yml
```diff
@@ -16,3 +16,6 @@ galaxy_info:
         - 7
   categories:
     - system
+dependencies:
+  - role: epel
+    become: yes
```
armab: Using `become: yes` when including anything external is usually a bad sign.

The good practice is to `become: yes` for every task in a role where it's 100% needed.
Eg. we do privilege escalation per task (only when command is not executed without sudo), not per role.

roles/mongodb/meta/main.yml
```diff
@@ -16,3 +16,6 @@ galaxy_info:
         - 7
   categories:
     - system
+dependencies:
+  - role: epel
```
armab: Should have conditional check: Include `epel` only when `ansible_os_family == "RedHat"`.

roles/st2/tasks/config_auth.yml
armab: So here we install `epel-release` and still including `epel` role?

stackstorm.yml
```diff
@@ -7,7 +7,7 @@
     - postgresql
     - nginx
     - st2repos
-    - st2
     - st2mistral
+    - st2
```
armab: Let's keep the original order.

We install `st2` first and then `st2mistral`.

roles/st2repos/tasks/redhat.yml
Mierdin: Just to be clear, for the `st2` role, you're saying I should provide a var that controls the parameter passed there, defaulting to `latest`?

armab: No, just change to `state: installed` here, as `ansible-lint` suggests.
> [ANSIBLE0010] Package installs should not use latest

Sorry for TL;DR ^^ :smile:

roles/st2/tasks/main.yml
```diff
@@ -1,17 +1,25 @@
...
+
+- name: Install latest st2 package (stable)
```
armab: `(stable)` is not exactly what we install here.

Package repository (if you meant that) is controlled via `st2_pkg_repo`, see: https://github.com/StackStorm/ansible-st2#variables

So `(stable)` note could be removed

roles/st2/tasks/main.yml
```diff
@@ -1,17 +1,25 @@
...
+  package:
     name: st2
     state: latest
   when: st2_version == "latest"
   notify:
     - restart st2
   tags: skip_ansible_lint
 
-- name: Install latest st2 package
+- name: Install latest st2 package (not stable)
```
armab: Same as https://github.com/StackStorm/ansible-st2/pull/66/files#r97132662

`(not stable)` note could be removed.

We can write `name: Install pinned st2 package version` or similar explanation instead.

roles/st2repos/tasks/redhat.yml
Mierdin: So the problem I had with this specific line (not other places where I made this linting error) was that unless I updated ca-certificates, I'd get this error when `st2mistral` runs:

`Failure talking to yum: Cannot retrieve repository metadata (repomd.xml) for repository: StackStorm_stable. Please verify its path and try again`

I just checked, and setting this to `present` instead of `latest` causes the error to come back up, as the package is already installed, so it doesn't do anything.

armab: Makes sense for an exception :+1:

This looks like old `EL6` version according to http://unix.stackexchange.com/questions/21310/yum-cannot-retrieve-repository-centos-6

If that happens on `EL6` only, can you add conditional check to run this task only on EL6?

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,22 @@
...
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
+    metadata_expire: 300
+    gpgcheck: no
+    enabled: yes
+    sslverify: yes
+
+# Fixes "Failure talking to yum: Cannot retrieve repository metadata (repomd.xml) for repository: StackStorm_stable. Please verify its path and try again" when installing st2
+- name: Update ca-certificates package
+  become: yes
+  package:
```
armab: Also, since this block is part of `redhat.yml`, I think it's more expected to use native `yum` module for installation, instead of cross-platform `package` module.

roles/st2repos/tasks/redhat.yml
```diff
@@ -0,0 +1,22 @@
...
+    repo_gpgcheck: yes
+    gpgkey: "https://packagecloud.io/StackStorm/{{ st2_pkg_repo }}/gpgkey"
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
+    metadata_expire: 300
+    gpgcheck: no
+    enabled: yes
+    sslverify: yes
+
+# Fixes "Failure talking to yum: Cannot retrieve repository metadata (repomd.xml) for repository: StackStorm_stable. Please verify its path and try again" when installing st2
+- name: Update ca-certificates package
```
armab: Since this fixes specific error for PackageCloud repository, maybe it's worth to move it _before_ adding st2 packagecloud repo?

roles/st2repos/tasks/redhat.yml
Mierdin: I may have a better fix for this, give me a few minutes.

Mierdin: @armab Okay, this seems to work: 9c52c7b Thoughts?

armab: @Mierdin It's not clear how https://github.com/StackStorm/ansible-st2/commit/9c52c7baa597c4645db3f98ed2742f76ac5bb2b8 fixes the issue with outdated `ca-certificates`.

According to http://stackoverflow.com/questions/26734777/yum-error-cannot-retrieve-metalink-for-repository-epel-please-verify-its-path and http://unix.stackexchange.com/questions/21310/yum-cannot-retrieve-repository-centos-6 the 2 possible fixes for outdated `EL6` are: 
- use `http` instead of `https` for package repo
- update `ca-certificates` package

How `yum -y repolist` fixes that? I'm probably don't have enough context here.
Additionally, can you reproduce this issue? As I understood it happens in some outdated `EL 6.5` distro.

I'd say, - we don't support some outdated RHEL6.5 which was released in `2013`.

So after all, - I'm +1 to remove this block completely :) WDYT?

roles/st2/tasks/main.yml
Mierdin: Fixed in https://github.com/StackStorm/ansible-st2/pull/66/commits/7fa3a70832051ce26dc981304beb20ca9db6b069

roles/st2/tasks/main.yml
Mierdin: I tried that before, got this:

```
[root@ip-10-0-4-91 ~]# subscription-manager repos --enable rhel-6-server-optional-rpms
Error: rhel-6-server-optional-rpms is not a valid repository ID. Use --list option to see valid repositories.
[root@ip-10-0-4-91 ~]# subscription-manager repos --list
This system has no repositories available through subscriptions.
```

Mierdin: Additionally, this task does run idempotently, now that I've changed to `state: latest`:

```
TASK [st2 : Install libffi-devel on EL6] ***************************************
ok: [10.0.4.91]
```

This should also make the `ignore_errors` statement pointless, so I have removed it in 100ca71. So as long as you don't prefer the `server-optional` approach (and know what I'm doing wrong that is preventing me from installing `libffi-devel` that way) then this should work better.

roles/mongodb/meta/main.yml
Mierdin: Ack, fixed in 9988842

roles/mongodb/meta/main.yml
Mierdin: Fixed in 9988842

roles/st2/tasks/main.yml
armab: +1
Thanks for explanation!

stackstorm.yml
Mierdin: Fixed in 56ed120

roles/st2/tasks/main.yml
Mierdin: Yeah, my intention with changing this was for easier debugging. Since both tasks had the same name, it was hard to tell from the output which one actually ran.

This should work better: 3c3f272

roles/st2/tasks/main.yml
Mierdin: Fixed in 3c3f272

roles/st2repos/tasks/redhat.yml
Mierdin: Okay, per our slack conversation, I reverted back to the old approach and am now updating `ca-certificates` using the `yum` module, in d8714bb

roles/st2repos/tasks/redhat.yml
Mierdin: Fixed in d8714bb

roles/st2repos/tasks/redhat.yml
Mierdin: Reordered in d8714bb

roles/st2/tasks/config_auth.yml
Mierdin: Fixed in e9ed597

Mierdin:
@armab ack your latest. Everything seems to work on EL6, and I addressed your two outstanding comments, so if those pass, this should be gtg. Will test on EL7 and CentOS6/7 and ubuntu, then merge latest `master`, and will let you know via slack when ready

PR #71, opened 2017-01-06
Updated kitchen and travis to support centos

### LINKED ISSUE #68: CentOS7 CI in Travis
We started adding `EL` support in https://github.com/StackStorm/ansible-st2/pull/65 and https://github.com/StackStorm/ansible-st2/pull/65 as part of https://github.com/StackStorm/ansible-st2/issues/5.
Working without `CI` is bad for quality, testing and reviewing.

There is a need to add `CentOS7` integration Testing in Travis as we already do for `Ubuntu Trusty` and `Ubuntu Xenial`. The framework is already setup.

It's possible to make `CentOS7` build non-voting to skip the failure for entire build (see [`travis.yml`](https://github.com/StackStorm/st2/blob/v2.1.0/.travis.yml#L34-L35) example for `st2` repo) until we re-write all roles to support new platforms.

See:
* https://travis-ci.org/StackStorm/ansible-st2
* https://github.com/StackStorm/ansible-st2/blob/master/.travis.yml
* https://github.com/StackStorm/ansible-st2/blob/master/.kitchen.yml

cc @Mierdin

Mierdin:
Closes #68 and #69


.kitchen.yml
```diff
@@ -33,6 +33,18 @@ platforms:
...
         - /sys/fs/cgroup:/sys/fs/cgroup:ro
+  - name: centos-6
+    driver_config:
+      image: stackstorm/packagingtest:centos6
+      platform: rhel
+      run_command: /sbin/init
+  - name: centos-7
+    driver_config:
+      image: stackstorm/packagingtest:centos7
+      platform: rhel
```
armab: I double-checked, you were right.
There is `centos` option according to https://github.com/test-kitchen/kitchen-docker#platform

Mierdin:
Cool. I also installed test kitchen on my laptop last night so I am not
flying blind any more

armab:
I investigated the build failure.
That's because we generate `/etc/ssh/ssh_host_rsa_key` already in our `Dockerfile`: https://github.com/StackStorm/st2-dockerfiles/blob/master/packagingtest/centos7/systemd/Dockerfile#L63 (and [here's](http://ask.xmodulo.com/sshd-error-could-not-load-host-key.html) why it's needed).

I've submitted patch in `kitchen-docker` https://github.com/test-kitchen/kitchen-docker/pull/253, but in a meantime you can try to use custom intermediate `Dockerfile` for `centos6` and `centos7` for `kitchen-docker` to work, see:
- https://github.com/test-kitchen/kitchen-docker#dockerfile
- https://github.com/test-kitchen/kitchen-docker/blob/master/test/Dockerfile
- https://github.com/test-kitchen/kitchen-docker/blob/b2b44bacb39673722454cf35762fd7b76037b80d/lib/kitchen/driver/docker.rb#L183-L277

Hope it'll unblock you.

Mierdin:
Yup that's the plan. Thanks

Mierdin:
@armab Should I also propose a change to https://github.com/StackStorm/st2-dockerfiles?

Dockerfile
```diff
@@ -0,0 +1,27 @@
+# This Dockerfile is a temporary solution so that we can test our centos
```
armab: It would be nice to hide these intermediate Dockerfiles under `.kitchen/` dir.
Just to avoid confusion from users when we place `Dockerfile` in the root dir of repo.

jvrplmlmn:
FWIW this issue https://github.com/StackStorm/chef-stackstorm/issues/30 in the [chef-stackstorm](https://github.com/StackStorm/chef-stackstorm) discusses why you shouldn't use the `stackstorm/packagingtest` Docker images for testing.

Mierdin:
Closing, as #81 and #82 solve this problem more correctly.

Issue #12, opened 2015-08-19
Publish on Ansible galaxy


armab:
Good "sign of quality" objective is to publish StackStorm roles on Ansible galaxy to get some more attention from the community.
At that point it should be tested/stable enough.
- [x] Register Ansible Galaxy account
- [ ] Polish, Refactor, Follow best practices
- [ ] Integration Tests #15 
- [ ] Ansible Galaxy Docs, Examples
- [ ] Decide: ship it as `stackstorm.st2` role or `st2`, `st2web`, `st2mistral`, `st2chatops` roles
- [x] ~Use Ansible Galaxy 3rd-party dependencies for `postgresql`, `rabbitmq`, `mongo`~ (We use our own roles now, but should leave an option for users to choose alternate external roles)

ytjohn:
Relates to:

https://github.com/StackStorm/ansible-st2/issues/95
https://github.com/StackStorm/ansible-st2/issues/45

ytjohn:
Under "Decide" I would recommend that you ship as "stackstorm.X". This lets you continue shipping your more common roles (nginx/postgresql/rabbitmq) under your name until you feel comfortable moving to any 3rd-party roles.

I would put that decision higher up, and register an account sooner rather than later. That way you can have the stackstorm username and stackstorm.<role name>, even if you don't list anything on galaxy.

Immediately after that, you should rename the roles in place. For example, if you register with the username "stackstorm", you would rename `roles/nginx` to `roles/stackstorm.nginx` and `roles/st2` to `roles/stackstorm.st2`.  Then you would update the playbook example to use the new names.

I would strongly encourage to do this step as early as possible. This gives those of us downloading the roles and dropping them into our `ext_roles` directory the ability to isolate them right away. Then, once you split the roles out into their own repos, they will already be named appropriately.

armab:
Sounds like a good plan and we're thinking alike because we already started tagging some roles in internal infra with `stackstorm.role-name` namespace :+1: 

So I just connected existing all-in-one `ansible-st2` repo to the Galaxy: 
https://galaxy.ansible.com/StackStorm/stackstorm/

Sadly `StackStorm` is case sensitive as is, - Galaxy takes it from the GitHub org name (see https://groups.google.com/forum/#!msg/ansible-project/RkBTwc7MIS4/E1ac8OFPAAAJ). I'd like `stackstorm.` namespace more, comparing to `StackStorm.`

[...]

PR #73, opened 2017-01-06
Ensure rabbitmq is enabled and running

armab:
Just a small addition to check that `rabbitmq-server` service is enabled and started which is usually a [good practice](http://docs.ansible.com/ansible/playbooks_best_practices.html#task-and-handler-organization-for-a-role) after installing some service.

The thing I missed in https://github.com/StackStorm/ansible-st2/pull/70

### DIFF
--- a/roles/rabbitmq/tasks/main.yml
+++ b/roles/rabbitmq/tasks/main.yml
@@ -1,9 +1,17 @@
 ---
-- name: Install rabbitmq packages on {{ ansible_distribution }}
+- name: Install rabbitmq package on {{ ansible_distribution }}
   become: yes
   package:
     name: rabbitmq-server
     state: present
-  tags: [rabbitmq]
   notify:
     - restart rabbitmq
+  tags: rabbitmq
+
+- name: Ensure rabbitmq is enabled and running
+  become: yes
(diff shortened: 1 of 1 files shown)


armab:
BTW, looks like we're missing the same check for `st2` services, see:
https://github.com/StackStorm/ansible-st2/blob/master/roles/st2/tasks/main.yml

If you guys are interested in submitting another PR to fix that, - it would be nice :+1: 

> BTW similar approach in `chef-stackstorm` repo: https://github.com/StackStorm/chef-stackstorm/blob/master/recipes/_services.rb#L19

jvrplmlmn:
Would be ideal to have Integration tests in place to verify this kind of things (related https://github.com/StackStorm/ansible-st2/issues/15).

armab:
Yeah, integration tests would be nice, but that's different thing comparing to this PR.

Additionally, as explained before in https://github.com/StackStorm/ansible-st2/issues/15 we didn't decide yet whether it will be `Inspec` or any other Integration testing Framework.

It think we'll return to this question a bit later, after adding support for all platforms and before publishing `asible-st2` on Galaxy: #12

Issue #76, opened 2017-01-11
PostgreSQL Role [EL]


armab:
External dependency [ANXS.postgresql](https://galaxy.ansible.com/list#/roles/512) doesn't support EL platform family.
We need to write custom role to install `PostgreSQL` for Mistral.

- [x] Tag all tasks in role
- [x] Meta
- [ ] Get best practices from [`ANXS.postgresql`](https://github.com/ANXS/postgresql) code
- [ ] Ensure service is listening on `127.0.0.1` by default (configurable), see: https://github.com/StackStorm/ansible-st2/issues/75
- [x] EL support
  - [x] CentOS6
  - [x] RHEL6
  - [x] CentOS7
  - [x] RHEL7
- [x] Idempotence check
- [x] Update README

cc @humblearner based on your feedback

armab:
Implemented in https://github.com/StackStorm/ansible-st2/issues/78 by @humblearner

Issue #11, opened 2015-08-19
Refactor Mistral Role [EL]


armab:
- [x] Rename `mistral` to `st2mistral`
- [x] Tag all tasks in role
- [x] Update meta file
- [x] EL support
  - [x] CentOS6
  - [x] RHEL6
  - [x] CentOS7
  - [x] RHEL7
- [x] Idempotence check

Depends on [PostgreSQL Role [EL] #76](https://github.com/StackStorm/ansible-st2/issues/76)

PR #78, opened 2017-01-11
Rhel support for mistral and create postgresql role (remove AXNS.postgresql dependency)

humblearner:
- [x]  Xenial
- [x]  Trusty
- [x]  Centos6
- [x]   Centos7
- [x]   Rhel6
- [x]   Rhel7

Covers https://github.com/StackStorm/ansible-st2/issues/76 and https://github.com/StackStorm/ansible-st2/issues/11


roles/mistral/tasks/main.yml
```diff
@@ -1,18 +1,19 @@
 ---
-
 - name: Install latest st2mistral
   become: yes
-  apt: 
+  package:
     name: st2mistral
     state: latest
```
armab: Yeah, it's good, considering condition in this block.
The idea here to install `latest` mistral eg. get upgrades.

roles/postgresql/tasks/postgresql_apt.yml
```diff
@@ -0,0 +1,14 @@
...
+  apt:
+    name: postgresql
+    state: installed
+  tags: [db, postgresql]
+
+- name: apt | Start and Enable Postgresql
+  become: yes
+  service:
+    name: postgresql
+    state: started
```
armab: action enabling service startup is missing

roles/postgresql/tasks/postgresql_yum6.yml
```diff
@@ -0,0 +1,53 @@
...
+    update_cache: yes
+  when: ansible_distribution == "CentOS"
+  tags: [db, postgresql]
+
+- name: yum | Get pgdg-centos94-9.4-2 for {{ ansible_distribution }}
+  become: yes
+  yum:
+    name: http://yum.postgresql.org/9.4/redhat/rhel-6-x86_64/pgdg-redhat94-9.4-2.noarch.rpm
+    state: installed
+    update_cache: yes
```
armab: `update_cache: yes` will break Ansible idempotence. Avoid using where possible

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,34 @@
...
+  tags: [db, postgresql]
+
+- name: yum | Initialize Postgresql
+  become: yes
+  command: postgresql-setup initdb
+  tags: [db, postgresql]
+
+- name: yum | MD5-encrypted password for postgres 1
+  become: yes
+  command: 'sed -i "s/\(host.*all.*all.*127.0.0.1\/32.*\)ident/\1md5/" /var/lib/pgsql/data/pg_hba.conf'
```
armab: Use http://docs.ansible.com/ansible/lineinfile_module.html instead

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,34 @@
...
+  tags: [db, postgresql]
+
+- name: yum | MD5-encrypted password for postgres 2
+  become: yes
+  command: 'sed -i "s/\(host.*all.*all.*::1\/128.*\)ident/\1md5/" /var/lib/pgsql/data/pg_hba.conf'
```
armab: Same, Ansibe module http://docs.ansible.com/ansible/lineinfile_module.html can help to make sure the command is Ansible-idempotent

stackstorm.yml
```diff
@@ -5,6 +5,7 @@
     - mongodb
     - rabbitmq
     - st2repos
+    - postgresql
     - mistral
```
armab: Please don't forget to rename `misral` to `st2mistral` as requested in https://github.com/StackStorm/ansible-st2/issues/11

So we'll use the following roles responsible for StackStorm:
- `st2`
- `st2mistral`
- `st2web`
- `st2chatops` in future

for consistency.

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,34 @@
...
+  tags: [db, postgresql]
+
+- name: yum | Start and Enable Postgresql
+  become: yes
+  service:
+    name: postgresql
+    state: started
+  tags: [db, postgresql]
```
armab: Additional "enable" action is missing

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,34 @@
...
+    update_cache: yes
+  with_items:
+    - postgresql-server
+    - postgresql-contrib
+    - postgresql-devel
+  tags: [db, postgresql]
+
+- name: yum | Initialize Postgresql
+  become: yes
+  command: postgresql-setup initdb
```
armab: This command needs idempotence, maybe `creates` some lock file, don't know what else we can do.

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: yum | Initialize postgresql
+  become: yes
+  yum:
+    name: "{{ item }}"
+    state: installed
+    update_cache: yes
```
armab: Yeah, avoid `update_cache` everywhere where it's possible.

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: yum | Initialize postgresql
```
armab: Probably you meant `Install postgresql`, because next block is `Initialize postgresql`

requirements.yml
```diff
@@ -1,7 +0,0 @@
----
-- hosts: all
-  connection: local
-  tasks:
-  - name: Install galaxy dependencies
-    command: ansible-galaxy install -r roles/mistral/requirements.yml
```
armab: :tada:

roles/postgresql/tasks/postgresql_yum6.yml
```diff
@@ -0,0 +1,53 @@
...
+- name: yum |  Get pgdg-centos94-9.4-2 for {{ ansible_distribution }}
+  become: yes
+  yum:
+    name: http://yum.postgresql.org/9.4/redhat/rhel-6-x86_64/pgdg-centos94-9.4-2.noarch.rpm
+    state: installed
+    update_cache: yes
+  when: ansible_distribution == "CentOS"
+  tags: [db, postgresql]
+
+- name: yum | Get pgdg-centos94-9.4-2 for {{ ansible_distribution }}
```
armab: This might be confusing. Probably it means "Setup PostgreSQL repository" or something like that.

roles/postgresql/tasks/postgresql_yum6.yml
```diff
@@ -0,0 +1,53 @@
...
+  become: yes
+  yum:
+    name: http://yum.postgresql.org/9.4/redhat/rhel-6-x86_64/pgdg-redhat94-9.4-2.noarch.rpm
```
armab: Can you check if `https://` version will work here (and in previous block)?  for yum install from url

roles/postgresql/tasks/postgresql_apt.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: apt | Initialize postgresql
```
armab: Probably `Install PostgreSQL` and so `Initialize PostgreSQL` step is missing.

README.md
```diff
@@ -26,8 +26,6 @@ ansible-playbook stackstorm.yml
 ```
 
 ## Dependencies
-Ansible Galaxy roles used by StackStorm installation:
```
armab: Please also remove more `requirements.yml` occurencies in README.md.

roles/postgresql/meta/main.yaml
```diff
@@ -0,0 +1,18 @@
...
+  license: Apache
+  min_ansible_version: 2.2
+  tags: [db, postgresql]
+  platforms:
+    - name: Ubuntu
+      versions:
+        - trusty
+        - xenial
+    - name: EL
+        - 6
```
armab: `versions:` is missing

roles/st2mistral/tasks/main.yml
```diff
@@ -1,18 +1,27 @@
 ---
+- name: Install st2mistral dependency
+  become: yes
+  package:
+    name: st2python
```
armab: Good catch :+1: 
I assume that's the case why mistral didn't work for you.

We install `st2python` as package dependency for `st2`, see: https://github.com/StackStorm/st2-packages/blob/a46cc2267ac4b23c8f21a3471824c7c7f430cd1a/packages/st2/rpm/st2.spec#L15

Seems like we should add the same dependency for `st2mistral` as well in `st2-packages`.

roles/st2mistral/tasks/main.yml
```diff
@@ -1,18 +1,27 @@
...
+    state: present
+  when: ansible_distribution == "CentOS" and ansible_distribution_major_version == "6"
```
armab: > `ansible_distribution == "CentOS"`
Will it work on `RHEL6`?

roles/postgresql/tasks/postgresql_yum6.yml
```diff
@@ -0,0 +1,65 @@
...
+  with_items:
+    - postgresql94-server
+    - postgresql94-contrib
+    - postgresql94-devel
+  tags: [db, postgresql]
+
+- name: yum | Initialize Postgresql-9.4
+  become: yes
+  command: service postgresql-9.4 initdb
+  tags: [db, postgresql]
```
armab: Same, this command needs a "file lock" to ensure that we run `initdb` only once per system and don't repeat it.
_OR_ it's possible to run `init` only when previous block "changed", eg. only after install.

See: https://raymii.org/s/tutorials/Ansible_-_Only-do-something-if-another-action-changed.html

roles/postgresql/tasks/postgresql_yum6.yml
```diff
@@ -0,0 +1,65 @@
...
+    name: libselinux-python
+    state: installed
+  tags: [db, postgresql]
+
+- name: yum | Install postgresql94
+  become: yes
+  yum:
+    name: "{{ item }}"
+    state: installed
+    update_cache: yes
```
armab: Can we avoid `update_cache` here to make this command Ansible-idempotent friendly?

With update_cache it will always have "changed" state.

roles/postgresql/tasks/postgresql_yum7.yml
armab: See another comment: https://github.com/StackStorm/ansible-st2/pull/78/files#r95995753 about executing an action only when another command is changed

roles/st2mistral/meta/main.yml
```diff
@@ -0,0 +1,21 @@
...
+  license: Apache
+  min_ansible_version: 2.2
+  tags: st2mistral
+  platforms:
+    - name: Ubuntu
+      versions:
+        - trusty
+        - xenial
+    - name: EL
+        - 6
```
armab: `versions:` is missing

roles/st2mistral/tasks/main.yml
armab: Created an Issue for that: https://github.com/StackStorm/st2-packages/issues/407 so we won't forget about it.

armab:
Overall looks nice :+1: 

The final thing I would like you to address is Ansible Idempotence.
Eg. Ansible-playbook re-run should end with the following results: 
```
changed=0.*failed=0
```

roles/postgresql/tasks/postgresql_yum7.yml
```diff
@@ -0,0 +1,42 @@
...
+    - postgresql-server
+    - postgresql-contrib
+    - postgresql-devel
+  register: install_postgresql
+  tags: [db, postgresql]
+
+- name: yum | Initialize Postgresql
+  become: yes
+  command: postgresql-setup initdb
+  when: install_postgresql.changed
```
armab: Nice indeed :+1:

README.md
```diff
@@ -19,15 +19,11 @@ At least 2GB of memory and 3.5GB of disk space is required, since StackStorm is
 
 ## Installation
 ```sh
-# ansible galaxy roles
-ansible-playbook requirements.yml
 # stackstorm
 ansible-playbook stackstorm.yml
 ```
 
 ## Dependencies
```
armab: This line could be removed as well, take a look at resulting `README.md`.

armab:
`ANXS.postgresql` dependency started to fail for previously successful `master` builds: 
```sh
TASK [ANXS.postgresql : PostgreSQL | Add PostgreSQL repository] ****************

       task path: /tmp/kitchen/roles/ANXS.postgresql/tasks/install.yml:17

       fatal: [localhost]: FAILED! => {"failed": true, "msg": "The conditional check 'postgresql_apt_repository' failed. The error was: expected token 'end of statement block', got 'http'\n  line 1\n\nThe error appears to have been in '/tmp/kitchen/roles/ANXS.postgresql/tasks/install.yml': line 17, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: PostgreSQL | Add PostgreSQL repository\n  ^ here\n"}
```
https://travis-ci.org/StackStorm/ansible-st2/jobs/191748821

So this PR is just in time!

roles/st2mistral/tasks/main.yml
humblearner: In the bootstrap scripts we install `st2` and its dependencies before `st2mistral`, that is why it was never caught: https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el6.sh#L653. However, we may not have another use cases besides this and docker, where anyone installs `st2mistral` before `st2`.

PR #80, opened 2017-01-13
Include meta for Galaxy install

armab:
Meta which allows installing `ansible-st2` repo via ansible-galaxy

```sh
ansible-galaxy install ...
```
And while `ansible-st2` is not a real single "role", since we store many roles in `roles/`, this workaround for `ansible.cfg` to find stackstorm roles in plays should work:
```ini
[defaults]
roles_path = /etc/ansible/roles/:/etc/ansible/roles/stackstorm/roles/
```


Mierdin:
LGTM 👍  - I assume I should update this in https://github.com/StackStorm/ansible-st2/pull/66 to include CentOS then?

PR #81, opened 2017-01-16
Enable CI for CentOS6

armab:
> Implements https://github.com/StackStorm/ansible-st2/issues/69

Enable CI for `CentOS6` (`kitchen-docker` in TravisCI).

Additionally, we run the task with `allow_failures` in Travis until CentOS6 is fully supported by `ansible-st2` playbooks as part of https://github.com/StackStorm/ansible-st2/milestone/2 plan.

Issue #38, opened 2016-01-13
Synchronize with recent Ansible


armab:
Fix the builds (since: https://circleci.com/gh/StackStorm/ansible-st2/31), check what's new in recent Ansible `2.0` version which broke the build, verify/test entire playbook story, check the backward compatibility.

PR #82, opened 2017-01-16
Enable CentOS7 CI in Travis

armab:
This implements #68 

Enable CI for `CentOS7` (`kitchen-docker` in TravisCI).

Allow CentOS7 build to fail in Travis until we fully implement platform support in Ansible playbooks as part of https://github.com/StackStorm/ansible-st2/milestone/2 plan.

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -38,6 +38,14 @@ platforms:
       platform: centos
       dockerfile: .kitchen-docker/centos6/Dockerfile
       run_command: /usr/sbin/sshd -D
+  # CentOS7 with Systemd
+  - name: centos-7
+    driver_config:
+      platform: centos
+      dockerfile: .kitchen-docker/centos7/Dockerfile
+      run_command: /sbin/init
+      volume:
+        - /sys/fs/cgroup:/sys/fs/cgroup:ro
 
 suites:
   - name: default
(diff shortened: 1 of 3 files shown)


.kitchen.yml
```diff
@@ -33,6 +33,17 @@ platforms:
...
       volume:
         - /sys/fs/cgroup:/sys/fs/cgroup:ro
+  # CentOS7 with Systemd
+  - name: centos-7
+    driver_config:
+      platform: centos
+      dockerfile: .kitchen-docker/centos7/Dockerfile
+      # TODO: Use non-root user after updating upstream Docker image: https://github.com/StackStorm/st2-dockerfiles/issues/38
+      username: root
+      password: docker.io
```
armab: As a workaround for CentOS7 we use different approach comparing to CentOS6 #81

As @Mierdin discovered before non-root login is blocked on current CentOS7 image. But we still can use root.

Additionally, I created [CentOS7 packagingtest needs re-build #38](https://github.com/StackStorm/st2-dockerfiles/issues/38) with detailed description so we fix it in future.

armab:
So CentOS7 checks are running, SSH works, Ansible is installed and playbooks is executed: https://travis-ci.org/StackStorm/ansible-st2/jobs/192437248

CI started to fail on all existing platforms (incl. `master` builds) due to external dependency failure in `ANXS.postgresql`, which is unrelated to this PR.

armab:
Rebased on top of master which removes external role `ANXS.postgresql` dependency and the build is back green.

PR #87, opened 2017-01-18
Rhel st2web support and nginx role

humblearner:
- [x] centos6
- [x]  centos7
- [x]  xenial
- [x]  trusty
- [x]  rhel6 (aws)
- [x]  rhel7


roles/st2web/tasks/st2web_apt.yml
```diff
@@ -0,0 +1,82 @@
+---
+- name: Install latest st2web package
+  become: yes
+  apt:
+    name: st2web
+    state: latest
+  when: st2_version == "stable"
```
armab: I just merged change #83, so we can use `latest` here as well.

roles/st2web/tasks/main.yml
```diff
@@ -1,94 +1,5 @@
...
-- name: Enable site
-  become: yes
-  file:
-    state: link
-    src: /etc/nginx/sites-available/st2.conf
-    dest: /etc/nginx/sites-enabled/st2.conf
-  notify:
-    - Restart nginx
+- name: Install st2web on {{ ansible_distribution }}
+  include: st2web_{{ ansible_pkg_mgr }}.yml
```
armab: Looks like a lot of code duplication here.
No sense for splitting, 2 files are differing with just a couple of small things at the moment.

Your previous approach with `package` module would be more elegant, see: https://github.com/StackStorm/ansible-st2/blob/32566695905cb92a4a975eba01e564702554b2fd/roles/st2mistral/tasks/main.yml#L12

roles/nginx/tasks/nginx_apt.yml
```diff
@@ -11,7 +9,15 @@
 - name: Add nginx repos
   become: yes
   apt_repository:
-    repo: "deb http://nginx.org/packages/ubuntu/ trusty nginx"
+    repo: "deb http://nginx.org/packages/ubuntu/ {{ ansible_distribution_release|lower }} nginx"
+    state: present
+    update_cache: yes
+  tags: nginx
+
+- name: Add nginx deb-src
```
armab: Entire block `deb-src` could be removed for `nginx`. We don't need package sources for sure.
Probably something we forgot to clean before.

roles/st2web/meta/main.yml
```diff
@@ -0,0 +1,23 @@
...
+      versions:
+        - trusty
+        - xenial
+    - name: EL
+      versions:
+        - 6
+        - 7
+  categories:
+    - system
+dependencies:
```
armab: So did you check this behavior?

If you remove `nginx` role from the `stackstorm.yml`, will `st2web` run `nginx` as a dependency?
Would like to be sure here.

armab: `dependencies` works OK, I see this in `ops-infra` testing nodes.
So nevermind about double checking.

roles/nginx/tasks/nginx_apt.yml
```diff
@@ -0,0 +1,30 @@
...
+    url: http://nginx.org/keys/nginx_signing.key
+    state: present
+  tags: nginx
+
+- name: Add nginx repos
+  become: yes
+  apt_repository:
+    repo: "deb http://nginx.org/packages/ubuntu/ {{ ansible_distribution_release|lower }} nginx"
+    state: present
+    update_cache: yes
```
Mierdin: Are we not doing this anymore?

roles/nginx/tasks/nginx_apt.yml
```diff
@@ -0,0 +1,30 @@
...
+  tags: nginx
+
+- name: Add nginx deb-src
+  become: yes
+  apt_repository:
+    repo: "deb-src http://nginx.org/packages/ubuntu/ {{ ansible_distribution_release|lower }} nginx"
+    state: present
+    update_cache: yes
```
Mierdin: Same comment as before

roles/nginx/tasks/nginx_apt.yml
armab: Per #90 we shouldn't `update_cache` for package installs, eg `apt` and `yum` modules, while this is `apt_repostiry` module.

For `apt_repository` module `update_cache: yes` is the default behavior. Optionally this line could be removed to avoid confusion with `apt`/`yum`.


So again: we do `update_cache: yes` when adding new repo (default) and we don't `update_cache: yes` when installing a package (default).

PR #92, opened 2017-01-20
Add libselinux-python and libsemanage-python dependencies for nginx

humblearner:


roles/nginx/tasks/nginx_yum.yml
```diff
@@ -1,3 +1,18 @@
+---
+- name: Dependency for nginx | libsemanage-python
```
Mierdin: Just a little nit - why not put these into a single task using a list?

humblearner: Sure, change on the way.

roles/nginx/tasks/nginx_yum.yml
```diff
@@ -1,3 +1,11 @@
+---
+- name: Install dependencies for nginx
```
armab: I believe it's not dependencies for nginx, but dependencies for http://docs.ansible.com/ansible/seboolean_module.html module to work.

See: https://github.com/StackStorm/ansible-st2/pull/92/files#diff-84ef76b67429d6f947d82255e8a3a5e5R36

armab: Will adjust in https://github.com/StackStorm/ansible-st2/pull/93

PR #93, opened 2017-01-20
Fix Selinux module work for nginx role

armab:
Fixes the following error under `EL 6/7`: 
```yaml
TASK [nginx : Allow network access for nginx] **********************************
task path: /tmp/kitchen/roles/nginx/tasks/nginx_yum.yml:34
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "SELinux is disabled on this host."}
```

Also includes other cosmetic changes for `nginx` role.

### DIFF
--- a/roles/nginx/tasks/nginx_yum.yml
+++ b/roles/nginx/tasks/nginx_yum.yml
@@ -37,4 +37,5 @@
     name: httpd_can_network_connect
     state: yes
     persistent: yes
+  when: ansible_selinux.status == "enabled" and ansible_selinux.mode == "enforcing"
   tags: nginx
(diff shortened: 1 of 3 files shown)


roles/nginx/tasks/nginx_yum.yml
```diff
@@ -31,10 +24,18 @@
...
+    state: present
+  tags: nginx
+
+- name: Adjust SELinux to allow network access for nginx
   become: yes
   seboolean:
     name: httpd_can_network_connect
     state: yes
     persistent: yes
+  when: ansible_selinux.status == "enabled" and ansible_selinux.mode == "enforcing"
```
armab: That's actual fix

PR #94, opened 2017-01-20
Rename st2mistral vars & Misc cosmetic improvements

armab:
We renamed `mistral` -> `st2mistral` role for consistency with other packages we install via roles (`st2`, `st2web`), but forgot to do so for respective variables.

This PR is a follow-up for https://github.com/StackStorm/ansible-st2/pull/78 which actualizes variable names with a role name (it's good practice to prefix variable name with a role name).

- `mistral_version` -> `st2mistral_version`
- `mistral_db` -> `st2mistral_db`
- `mistral_db_username` -> `st2mistral_db_username`
- `mistral_db_password` -> `st2mistral_db_password`

Additionally tiny improvements:
- Cosmetic fixes, more clear Task descriptions
- Fix some st2mistral DB migration corner cases with "file locks" we create
- Ensure `mistral` service is enabled and running


armab:
So after playing a bit and taking closer look at `st2mistral` role, I created a new issue to improve it in some better times: 
> [Move DB creation logic from st2mistral to postgresql role #95](https://github.com/StackStorm/ansible-st2/issues/95)

roles/st2mistral/tasks/main.yml
```diff
@@ -40,11 +29,9 @@
...
   register: mistral_deploy_database_init_script
-  notify:
-    - restart mistral
   tags: st2mistral
 
-- name: Initiate database
+- name: Initiate mistral database
   become: yes
   become_user: postgres
   shell: psql < /etc/mistral/init_mistral_db.SQL
```
armab: Here we have a problem when user considering to change any of `st2mistral_` database settings in future, - the `SQL` script will not be run twice. That means, DB is not created.

We can workaround with additional lock files, but https://github.com/StackStorm/ansible-st2/issues/95 should solve it in a right way.

roles/st2mistral/tasks/main.yml
```diff
@@ -55,9 +42,29 @@
     - restart mistral
   tags: st2mistral
 
-- name: Setup Mistral DB tables, etc
+- name: Make sure "Initiate mistral database" doesn't run twice
```
Mierdin: You're creating the file here, but I don't see where you're checking for the presence of the file in the previous task (i.e. no `when` clause)

roles/st2mistral/tasks/main.yml
```diff
@@ -55,9 +42,29 @@
...
+    dest: /etc/mistral/mistral.conf
+    section: database
+    option: connection
+    value: postgresql://{{ st2mistral_db_username }}:{{ st2mistral_db_password }}@localhost/{{ st2mistral_db }}
+    backup: yes
+  notify:
+    - restart mistral
+  tags: st2mistral
+
+- name: Setup mistral DB tables, etc
```
Mierdin: Same as above - don't currently see how this is idempotent, since there is no `when` clause

roles/st2mistral/tasks/main.yml
```diff
@@ -67,11 +74,19 @@
 
 - name: Register mistral actions
   become: yes
-  shell: /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate && touch /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
+  shell: /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf populate && touch /etc/mistral/mistral-db-manage.populate.ansible.has.run
```
Mierdin: same

roles/st2mistral/tasks/main.yml
armab: See: `creates: /etc/mistral/init_mistral_db.SQL.ansible.has.run` in previous task.
If file from the `creates` exists, - the command won't be executed.

See: http://docs.ansible.com/ansible/shell_module.html
for more info

roles/st2mistral/tasks/main.yml
armab: We touch the file first, find: `&& touch /etc/mistral/mistral-db-manage.populate.ansible.has.run`
and later in block don't execute it if `touch` exists: 
```yaml
creates: /etc/mistral/mistral-db-manage.populate.ansible.has.run`
```

roles/st2mistral/tasks/main.yml
armab: Similar example with `creates`: https://github.com/StackStorm/ansible-st2/pull/94/files/2b9a44658672559fff187610578832c22a6c2a0c#r97388431

roles/st2mistral/tasks/main.yml
Mierdin: Ah, okay I get it now.

roles/st2mistral/tasks/main.yml
```diff
@@ -1,37 +1,26 @@
...
+- name: Install pinned st2mistral package
   become: yes
   package:
-    name: st2mistral={{ mistral_version }}
+    name: st2mistral={{ st2mistral_version }}
     state: present
-  when: mistral_version != "latest"
-  tags: st2mistral
-
-- name: Configure mistral
```
humblearner: We don't need this task ?

armab:
@humblearner I just moved that  `mistral.conf` task after DB creation stage.

So DB is created first and then we change the `mistral.conf` setting if DB creation was successful.
Logically, it looks a bit better order for me.

PR #97, opened 2017-01-23
Added step to disable requiretty

Mierdin:
The current `st2` role does not disable the "requiretty" option in `/etc/sudoers`, which is a required step for setting up stackstorm, and as a result, the `st2smoketest` role was failing when it was time to run st2 commands:

```
[ec2-user@ip-10-0-4-218 ~]$ st2 run core.local_sudo -- date -R
.
id: 58866ee4ddd8ff29343e2d1b
status: failed
parameters:
  cmd: date -R
result:
  failed: true
  return_code: 1
  stderr: 'sudo: sorry, you must have a tty to run sudo'
  stdout: ''
  succeeded: false
[ec2-user@ip-10-0-4-218 ~]$
```

This PR introduces this step in the configuration.

### DIFF
--- a/roles/st2/tasks/user.yml
+++ b/roles/st2/tasks/user.yml
@@ -33,6 +33,14 @@
     validate: 'visudo -cf %s'
   tags: [st2, user]
 
+- name: Disable requiretty
+  become: yes
+  replace:
+    dest: "/etc/sudoers"
+    regexp: '^Defaults\s+\+?requiretty'
+    replace: '# Defaults requiretty'
+  when: st2_system_user_in_sudoers
+
 - name: Configure system user in /etc/st2/st2.conf
   become: yes
   ini_file:
@@ -50,10 +58,3 @@
     option: ssh_key_file
     value: "{{ _user.ssh_key_file }}"
     backup: yes
-
-- name: Disable requiretty
-  become: yes
-  replace:
-    dest: "/etc/sudoers"
-    regexp: '^Defaults\s+\+?requiretty'
-    replace: '# Defaults requiretty'
--- a/roles/st2/tasks/user.yml
+++ b/roles/st2/tasks/user.yml
@@ -40,6 +40,7 @@
     regexp: '^Defaults\s+\+?requiretty'
     replace: '# Defaults requiretty'
   when: st2_system_user_in_sudoers
+  tags: [st2, user]
 
 - name: Configure system user in /etc/st2/st2.conf
   become: yes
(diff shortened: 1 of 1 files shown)


roles/st2/tasks/user.yml
```diff
@@ -51,3 +50,10 @@
     option: ssh_key_file
     value: "{{ _user.ssh_key_file }}"
     backup: yes
+
+- name: Disable requiretty
+  become: yes
+  replace:
+    dest: "/etc/sudoers"
```
armab: From the http://docs.ansible.com/ansible/replace_module.html can we use "recommended" `path`, instead of old aliases: `dest, destfile and name` which could become deprecated in future.

roles/st2/tasks/user.yml
```diff
@@ -51,3 +50,10 @@
...
+    regexp: '^Defaults\s+\+?requiretty'
+    replace: '# Defaults requiretty'
```
armab: Additionally, please take a look at https://github.com/stackstorm/ansible-st2#variables `st2_system_user_in_sudoers` var and https://github.com/StackStorm/ansible-st2/blob/156a0844f83a37bd601420be2a10cc85b64bc34e/roles/st2/tasks/user.yml#L32 usage.

So we need to change `sudoers` only when `st2_system_user_in_sudoers: yes`

roles/st2/tasks/user.yml
```diff
@@ -51,3 +50,10 @@
     option: ssh_key_file
     value: "{{ _user.ssh_key_file }}"
     backup: yes
+
+- name: Disable requiretty
```
armab: Just a preference if we move this block after related task: `user | Add system user to sudoers`.

roles/st2/tasks/user.yml
Mierdin: How's this? e97a253

roles/st2/tasks/user.yml
```diff
@@ -33,6 +33,14 @@
     validate: 'visudo -cf %s'
   tags: [st2, user]
 
+- name: Disable requiretty
+  become: yes
+  replace:
+    dest: "/etc/sudoers"
+    regexp: '^Defaults\s+\+?requiretty'
+    replace: '# Defaults requiretty'
+  when: st2_system_user_in_sudoers
```
armab: Add missing tags please, see previous Tasks (all tagged).

armab:
:+1:

PR #101, opened 2017-01-25
Fix task include in postgresql role

armab:
Actual error:
```
FAILED! => {"failed": true, "reason": "the file_name '...roles/postgresql/tasks/postgresql_apt.yml' does not exist, or is not readable"}
```


armab:
Merging this hotfix fast, since I'm currently using & testing this set of stackstorm roles as dependency.

PR #104, opened 2017-01-28
Save credentials in .st2/config file

dzimine:
Save user credentials for `root` in `/root/.st2/config` as [installer script does](https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L397-L401)

It is convenient to have the system ready-to-use after install, but if one doesn't like it, turn off by `st2_save_credentials: no`.

### DIFF
--- a/roles/st2/tasks/auth.yml
+++ b/roles/st2/tasks/auth.yml
@@ -41,7 +41,10 @@
     - restart st2stream
 
 - name: auth | Create root's CLI configuration directory
-  file: path=/root/.st2 state=directory
+  become: yes
+  file:
+    path: /root/.st2
+    state: directory
   when: st2_save_credentials
 
 - name: auth | Save credentials in CLI configuration file
@@ -49,6 +52,9 @@
   blockinfile:
     dest: /root/.st2/config
     create: yes
+    mode: 0700
+    owner: root
+    group: root
     block: |
       [credentials]
       username = {{ st2_auth_username }}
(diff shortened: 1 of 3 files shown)


roles/st2/tasks/auth.yml
```diff
@@ -39,3 +39,18 @@
   notify:
     - restart st2api
     - restart st2stream
+
+- name: auth | Create root's CLI configuration directory
+  file: path=/root/.st2 state=directory
```
armab: This task needs `become: yes`

roles/st2/tasks/auth.yml
```diff
@@ -39,3 +39,18 @@
...
+
+- name: auth | Save credentials in CLI configuration file
+  become: yes
+  blockinfile:
+    dest: /root/.st2/config
+    create: yes
+    block: |
+      [credentials]
+      username = {{ st2_auth_username }}
+      password = {{ st2_auth_password }}
```
armab: For security reasons please set explicitly `mode` so the file with auth credentials will be readable only by `root` user.

In addition to that, just to be sure it would be nice to include `owner` and `group` settings as well.
https://docs.ansible.com/ansible/blockinfile_module.html

roles/st2/tasks/auth.yml
armab: Additionally, to be consistent with other playbooks in this repo, please use native multi-line yaml syntax (avoid `a=b`).
See: https://www.ansible.com/blog/ansible-best-practices-essentials for more info.

dzimine:
Thanks @armab , updated.

Issue #103, opened 2017-01-28
"Register mistral actions" fails on `staging-unstable`.


dzimine:
Installation of `staging-unstable` breaks on Ubuntu with the following exception:

```
Exception: Versioning for this project requires either an sdist tarball, or access to an upstream git repository. Are you sure that git is installed?
```
Yes, git **IS** installed on the box. 

After it happens, re-running the playbook skips the Register mistral actions step.
```
TASK [st2mistral : Register mistral actions] ***********************************
skipping: [st2]
```

Playbook:
```
---
- name: Install StackStorm
  hosts: st2
  strategy: debug
  vars:
    st2_pkg_repo: staging-unstable

  roles:
    - mongodb
    - rabbitmq
    - postgresql
    - st2repos
    - st2
    - st2mistral
    - nginx
    - st2web
    - st2smoketests
```

```
TASK [st2mistral : Register mistral actions] ***********************************
[...]

PR #105, opened 2017-01-30
Fix mistral register re-run on failure

armab:
> Partially fixes https://github.com/StackStorm/ansible-st2/issues/103#issuecomment-275851955

If `mistral-db-manage` failed first time, - it's skipped on next playbok re-run. This is wrong behavior.

This caused by blocking conditions in `when:`, - they are removed now.


humblearner:
`when` was to make the task: https://github.com/StackStorm/ansible-st2/pull/105/files#diff-6411b73900eb950467182a77f5e8f74aR73 idempotent. If it is still idempotent, 👍.

PR #109, opened 2017-01-31
Fix to circumvent pbr related bug in mistral-db-manage populate

m4dcoder:
Init of git is added here to temporarily fix a bug in python-muranoclient during pbr version checking on OpenStack action generations.


armab:
Fixes #103

PR #115, opened 2017-02-09
Update CentOS Docker images endpoints

armab:
Revert temporary workaround from https://github.com/StackStorm/ansible-st2/pull/82#discussion_r96282814 to use `root` in CentOS7 docker container.

It's possible since we're building `packagingtest` https://hub.docker.com/r/stackstorm/packagingtest/ Docker images automatically on every push to [st2-dockerfiles](https://github.com/StackStorm/st2-dockerfiles/tree/master/packagingtest) repo and can fix upstream images more easily.

See: https://github.com/StackStorm/st2-packages/pull/417 and https://github.com/StackStorm/st2-dockerfiles/issues/38 for more info

### DIFF
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -44,9 +44,6 @@ platforms:
     driver_config:
       platform: centos
       dockerfile: .kitchen-docker/centos7/Dockerfile
-      # TODO: Use non-root user after updating upstream Docker image: https://github.com/StackStorm/st2-dockerfiles/issues/38
-      username: root
-      password: docker.io
       run_command: /sbin/init
       volume:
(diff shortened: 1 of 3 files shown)


.kitchen.yml
```diff
@@ -44,9 +44,6 @@ platforms:
     driver_config:
       platform: centos
       dockerfile: .kitchen-docker/centos7/Dockerfile
-      # TODO: Use non-root user after updating upstream Docker image: https://github.com/StackStorm/st2-dockerfiles/issues/38
```
armab: Fixed by https://github.com/StackStorm/st2-dockerfiles/pull/43/files

Issue #74, opened 2017-01-09
Pack Install


armab:
Ability to list st2 packs to install as well as provide configurations/settings for them.

https://docs.stackstorm.com/pack_configs.html

Alternative example: https://github.com/stackstorm/puppet-st2#installing-and-configuring-packs

armab:
One of the possible syntaxes for the new interface with the pack install/configure: 

```yaml
st2_packs:
  # just `st2 pack install ansible`
  - name: ansible

  - name: aws
    # version pinning
    version: 1.2.3
    # state `present`/`absent`
    state: present
    # in-line config
    config:
      setup:
        region: "us-west-2"
        aws_access_key_id: ""
        aws_secret_access_key: ""
      interval: 20

  - name: github
    config:
      token: "abc123def"
      user: ""
      password: ""
      repository_sensor:
        repositories:
          - user: StackStorm
            name: st2

   # alternative way to use already prepared file which will be
   # transferred to /opt/stackstorm/configs/servicenow.yml
  - name: servicenow
    config: files/configs/servicenow.yml

  # uninstall pack
  - name: consul
    state: absent
```

mrlesmithjr:
FWIW, this is the method that I use currently which actually works pretty well for installing/uninstalling packs. I do agree, however, that there should be a better way. I was also recently looking at ways to leverage the JSON/YAML output in different ways.

https://github.com/mrlesmithjr/ansible-stackstorm/blob/master/tasks/packs.yml

armab:
Hi @mrlesmithjr, thanks for sharing!
Glad to see you here! I remember your StackStorm roles, which are great :+1:

BTW we recently reworked our entire pack installation + discovery story and introduced versioning: https://stackstorm.com/2017/01/31/stackstorm-exchange-new-pack-management/ with https://exchange.stackstorm.org/ and many other features.

I tried to play with just simple Ansible functionality to achive described above Ansible syntax, - but yaml logic is definitely not enough. Some of the problems we want to fix in a long term: 

- Pack install Idempotence, see: https://github.com/StackStorm/st2/issues/3052 (st2 core changes)
- Pack install version
  - if pack version unchanged, - we shouldn't try to re-install same pack again
  - if different version provided, - it should trigger reinstall in Ansible
- Variety of pack install methods/syntaxes (from the example above and https://docs.stackstorm.com/packs.html#installing-a-pack)
- Different sets of pack configurations

---

[...]

PR #117, opened 2017-02-13
Implement Pack Install & Configure

armab:
Implement logic described in https://github.com/StackStorm/ansible-st2/issues/74 to `install` & `configure` StackStorm integration packs from Ansible.


armab:
After trying more workarounds, it's not possible to make syntax described in https://github.com/StackStorm/ansible-st2/issues/74#issuecomment-277352903 with pure Ansible playbook/yaml logic.

This will require us to write custom Ansible module similar to `apt`, `yum`, `pip` which will work with `st2client` lib.

Issue #119, opened 2017-02-14
(feature request) Adding tasks to set up cluster configuration in the RabbitMQ role


userlocalhost:
This is a much-needed project for me. I want to build StackStorm in production environment with this.

The current tasks in the RabbitMQ role just only install package and start service.
But the RabbitMQ servers that I expect are set up a cluster configuration.

I want some tasks to set up cluster configuration in the RabbitMQ role.

Thank you.

armab:
Thanks for the feature request!
I agree, cluster and HA-friendly deployment is must have story for production use.

And while it may take a lot of time to implement it, - all dependent services we have like RabbitMQ, Mongo, PostgreSQL could have a lot of advanced configuration to handle.

But for now I'd suggest alternative approach.
`st2` role was written without hard dependency on internal `rabbitmq` role.
So instead of https://github.com/StackStorm/ansible-st2/issues/120 you can use external rabbitmq role
and [Configure with Existing Mongo, RabbitMQ, PostgreSQL installations #17](https://github.com/StackStorm/ansible-st2/issues/17)

All we need now is to add respective [`st2.conf` messaging setting](https://github.com/StackStorm/st2/blob/74d3dbdd3b996f1b97c9330c6c50d8de6370d41a/conf/st2.conf.sample#L153) so you could configure st2 to work with external rabbitmq cluster, deployed by another role.


Resulting play may look this way:
```yaml
---
- name: Configure RabbitMQ cluster
  hosts: rabbitmq-cluster
  roles:
    # See: https://github.com/alexey-medvedchikov/ansible-rabbitmq
    - alexeymedvedchikov.rabbitmq
      vars:
        rabbitmq_create_cluster: yes
        rabbitmq_erlang_cookie: st2-mq
        rabbitmq_users:
          - user: st2
            password: st2
            vhost: /
            configure_priv: .*
            read_priv: .*
            write_priv: .*
            tags: st2

- name: Install StackStorm
  hosts: st2
  roles:
    - name: Install and Configure `st2` itself
      role: st2
      vars:
        st2_messaging_url: "amqp://st2:st2@rabbitmq-cluster001:5672//"

    - st2mistral
    - st2web
    - st2smoketests
```

Just for reference I found a role which configures rabbitmq cluster https://github.com/alexey-medvedchikov/ansible-rabbitmq just to give an idea.

[...]

PR #120, opened 2017-02-15
Add tasks for setting up a rabbitmq cluster

userlocalhost:
This is the implementation for #119

This patch added a task that wrote shared value to the `.erlang.cookie` file for setting up a RabbitMQ cluster.
When a couple of RabbitMQ servers are deployed by this role, user can easily set it up by this patch.


userlocalhost:
I'll close it caused by [this discussion](https://github.com/StackStorm/ansible-st2/issues/119#issuecomment-280019176).

PR #122, opened 2017-02-22
add proxy option

ytjohn:
This adds the option for setting proxy environment variables whenever an http or https operation is planned (apt-key, rpm -i <url>). 

Example of running (I setup squid on localhost):

    root@ansible-st2-ubuntu14:/vagrant# apt-key del D59097AB
    OK
    root@ansible-st2-ubuntu14:/vagrant# ansible-playbook stackstorm.yml --extra-vars='{"st2_proxies": {"http_proxy": "http://localhost:3128", "https_proxy": "http://localhost:3128"}}' -i 'localhost,' --connection=local
    
    TASK [st2repos : Add keys to keyring] ******************************************
    changed: [localhost]

    PLAY RECAP *********************************************************************
    localhost                  : ok=54   changed=1    unreachable=0    failed=0

From /var/log/squid3/access.log 

    1487778697.480      4 127.0.0.1 TCP_MISS/200 3358 CONNECT localhost:443 - HIER_DIRECT/127.0.0.1 -
    1487778803.128    389 127.0.0.1 TCP_MISS/200 6090 CONNECT packagecloud.io:443 - HIER_DIRECT/50.97.198.58 -
    1487778803.606    476 127.0.0.1 TCP_MISS/200 6638 CONNECT packagecloud.io:443 - HIER_DIRECT/50.97.198.58 -
    1487778804.007    392 127.0.0.1 TCP_MISS/200 6090 CONNECT packagecloud.io:443 - HIER_DIRECT/50.97.198.58 -
    1487778804.394    376 127.0.0.1 TCP_MISS/200 6090 CONNECT packagecloud.io:443 - HIER_DIRECT/50.97.198.58 -
    1487778804.867    472 127.0.0.1 TCP_MISS/200 10296 CONNECT packagecloud.io:443 - HIER_DIRECT/50.97.198.58 -
    1487778950.596    555 127.0.0.1 TCP_MISS/200 6090 CONNECT packagecloud.io:443 - HIER_DIRECT/50.97.198.58 -
    
And to just validate nothing is broken:

    root@ansible-st2-ubuntu14:/vagrant# apt-key del D59097AB
    OK
    root@ansible-st2-ubuntu14:/vagrant# ansible-playbook stackstorm.yml -i 'localhost,' --connection=local
    
    TASK [st2repos : Add keys to keyring] ******************************************
    changed: [localhost]
    PLAY RECAP *********************************************************************
    localhost                  : ok=54   changed=1    unreachable=0    failed=0

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | ------------------------ | ------------- | ------------ |
 | **st2repo**
 | `st2_pkg_repo`           | `stable`      | StackStorm PackageCloud repository to install. [`stable`](https://packagecloud.io/StackStorm/stable/), [`unstable`](https://packagecloud.io/StackStorm/unstable/), [`staging-stable`](https://packagecloud.io/StackStorm/staging-stable/), [`staging-unstable`](https://packagecloud.io/StackStorm/staging-unstable/)
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
 | **st2**
 | `st2_version`            | `latest`      | StackStorm version to install. Use latest `latest` to get automatic updates or pin it to numeric version like `2.1.1`.
 | `st2_revision`           | `1`           | StackStorm revision to install. Used only with pinned `st2_version`.
@@ -42,11 +43,23 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2_auth_username`      | `testu`       | Username used by StackStorm standalone authentication.
 | `st2_auth_password`      | `testp`       | Password used by StackStorm standalone authentication.
 | `st2_save_credentials`   | `yes`         | Save credentials for local CLI in `/root/.st2/config` file.
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
 | **st2mistral**
 | `st2mistral_version`     | `latest`      | st2mistral version to install. Use latest `latest` to get automatic updates or pin it to numeric version like `2.1.1`.
 | `st2mistral_db`          | `mistral`     | PostgreSQL DB name for Mistral.
 | `st2mistral_db_username` | `mistral`     | PostgreSQL DB user for Mistral.
 | `st2mistral_db_password` | `StackStorm`  | PostgreSQL DB password for Mistral.
+| **epel**
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
+| **mongodb**
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
+| **nginx**
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
+| **nginx**
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
+| **postgresql**
+| `st2_proxies`             | `{}`         | Dictionary for http_proxy and https_proxy environment variables.
+
 
 ## Examples
 Install latest `stable` StackStorm with all its components on local machine:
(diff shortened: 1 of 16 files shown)


armab:
There is a simpler way.

See https://groups.google.com/forum/#!msg/ansible-project/abGxs8J_sII/vOARmHCqO6YJ discussion and https://github.com/ansible/ansible/pull/8651

So you can define environment in you play this way and it will be passed to all roles:

```yaml
---
- name: Install st2
  hosts: all
  environment:
    http_proxy: http://proxy.example.com:8080
  roles:
    - mongodb
    - rabbitmq
    - postgresql
    - st2repos
    - st2
    - st2mistral
    - nginx
    - st2web
    - st2smoketests
```

This is also recommended by Ansible docs: http://docs.ansible.com/ansible/playbooks_environment.html which demonstrates using `environment` at a play level and via `group_vars`.

-----

So thanks for the call, installing st2 via proxy is pretty popular setting indeed for our users.
I'd suggest to use the environment at a higher level (play) and just update `README.md` to demonstrate it :+1:

ytjohn:
@armab  Thank you for the discussion link. I did not see that when I ran into trying to install this on our network.

One issue: If you set your http(s) proxies globally for the playbook, then that will break the smoketests.

```
TASK [st2smoketests : get authentication token] ********************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": ["st2", "auth", "testu", "-p", "testp", "-t"], "delta": "0:00:00.391983", "end": "2017-02-22 18:00:32.459234", "failed": true, "rc": 1, "start": "2017-02-22 18:00:32.067251", "stderr": "", "stdout": "ERROR: 403 Client Error: Forbidden\nUnable to retrieve detailed message from the HTTP response. Expecting value: line 1 column 1 (char 0)\n for url: http://127.0.0.1:9100/tokens", "stdout_lines": ["ERROR: 403 Client Error: Forbidden", "Unable to retrieve detailed message from the HTTP response. Expecting value: line 1 column 1 (char 0)", " for url: http://127.0.0.1:9100/tokens"], "warnings": []}
	to retry, use: --limit @/vagrant/stackstorm.retry

```

Granted, this is the only place I know of in the playbook outside of apt-key/rpm that does a http request. So the documentation could be to leave st2smoketests out of your playbook. Or we could possibly unset the environment variables within the st2smoketests role or even individual tasks.

I felt targeting each task that specifically needs a proxy environment variable was a better long term approach, such as a future role that further customizes using api calls. 

Thoughts?

armab:
Setting proxy ENV per task is a PITA for sure.
Think about it globally in context of using external Ansible galaxy roles which are not controlled by us. Additionally, while adding proxy ENV per task we might introduce mistakes here and there during development. So setting proxy environment at a high level is the way to go.

To fix your issue, try setting `no_proxy="127.0.0.1,localhost"` environment variable with a list of hosts to exclude from proxying.

ytjohn:
Thank you for the feedback. I will close this one and hopefully open a PR with a documentation update.  

On a side note, I am glad to see a serious set of roles and playbooks for stackstorm. The documentation on the site really pushes the single install script over all the other methods. I have had mixed success with that script in previous attempts to try out stackstorm. I attempted to use it earlier this week on a vm in our lab environment (requiring a proxy) and that was a disaster. I had to essentially download the scripts it was downloading and modify them to include the environment variables. 

I would like to see that single install bash script replaced and become a wrapper around using the ansible playbook.  Foreman does something similar. They have a set of puppet modules used to install Foreman, and a "foreman-installer" which wraps around applying those puppet modules.

armab:
@ytjohn Thanks, that's useful

It's true, `curl | bash` script is mostly for demonstrational purposes which codifies manual steps described in https://docs.stackstorm.com/install/deb.html and can't handle any complex deployments.

As noted before, keep in mind that there are more issues when working with StackStorm via proxy (ex: pack install): https://github.com/StackStorm/st2/issues/3137 https://github.com/StackStorm/st2docs/issues/352 
We're definitely interested to improve that story in upcoming StackStorm releases.

In a meantime, please don't forget to add a doc PR if proxy worked for you!  :+1:

ytjohn:
PR #123

PR #124, opened 2017-03-04
Add nodejs role

humblearner:


roles/nodejs/tasks/nodejs_apt.yml
```diff
@@ -0,0 +1,31 @@
...
+  tags: nodejs
+
+- name: Add nodesource repos debs
+  become: yes
+  apt_repository:
+    repo: "{{ item }}"
+    state: present
+  with_items:
+    - "deb https://deb.nodesource.com/node_4.x {{ ansible_distribution_release }} main"
+    - "deb-src https://deb.nodesource.com/node_4.x {{ ansible_distribution_release }} main"
```
armab: We don't need source packages

roles/nodejs/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+# nodejs version to install
+nodejs_version: "4.x"
```
armab: I see this var is used in some places, in other places `4.x` is hardcoded. So decide or use it everywhere or hardcode it inside the tasks.

armab: From the one side `4.x` looks confusing and I'm afraid this could be used in a wrong way. Additionally, not sure if it's a good idea to allow configuring this var to install untested by us node version.

From the other side some configuration like version is not bad.

Another idea is `nodejs_major_version: 4` and use `{{nodejs_major_version}}.x` in code.

Decide what looks best for you.

roles/nodejs/meta/main.yml
```diff
@@ -0,0 +1,22 @@
...
+        - trusty
+        - xenial
+    - name: EL
+      versions:
+        - 6
+        - 7
+  categories:
+    - system
+  dependencies:
+    - role: epel
```
armab: Just a question: why do we need `epel` role for installing `nodejs`?

PR #125, opened 2017-03-04
Add st2chatops role

humblearner:


armab:
Installation looks pretty clean.

From the configuration point of view we'll need the following syntax: 
```yaml
st2chatops_version: latest
# hash to configure any value in st2chatops.env
# https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env
st2chatops_config:
  ST2_API_KEY: 123asd456
  HUBOT_ADAPTER: slack
  HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
  ...
```

humblearner:
Awesome!! That was my next question, whether we need configuration or not. Thanks!!

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  tags: [st2chatops, skip_ansible_lint]
+
+- name: Install pinned st2chatops package
+  become: yes
+  package:
+    name: st2chatops={{ st2chatops_version }}
+    state: present
+  when: st2chatops_version != "latest"
+  notify:
+    - restart mistral
```
lakshmi-kannan: Copy pasta

armab: Haha, right in the bull's-eye! :dart: :1st_place_medal:

humblearner: haha, WIP folks 😉

README.md
```diff
@@ -43,10 +43,15 @@ Below is the list of variables you can redefine in your playbook to customize st
...
 | `st2_save_credentials`   | `yes`         | Save credentials for local CLI in `/root/.st2/config` file.
 | **st2mistral**
-| `st2mistral_version`     | `latest`      | st2mistral version to install. Use latest `latest` to get automatic updates or pin it to numeric version like `2.1.1`.
+| `st2mistral_version`     | `latest`      | st2mistral version to install. Use latest `latest` to get automatic updates or pin it to numeric version like `2.2.0`.
 | `st2mistral_db`          | `mistral`     | PostgreSQL DB name for Mistral.
 | `st2mistral_db_username` | `mistral`     | PostgreSQL DB user for Mistral.
 | `st2mistral_db_password` | `StackStorm`  | PostgreSQL DB password for Mistral.
+| **st2chatops**               | Default hubot adapter: 'slack'
+| `st2chatops_version`     | `latest`      | st2mistral version to install. Use `latest` to get automatic updates or pin it to numeric version like `2.2.0`.
+| `st2chatops_ST2_API_KEY` | `CHANGE-ME-PLEASE`      | st2 API key, that can be generated using `st2 apikey create -k` in a task.
```
armab: The problem with this approach is that we need a new var for every setting in `st2chatops.env`. And if we introduce new setting (like recent Mattermost support https://github.com/StackStorm/st2chatops/pull/65/files) it will force us to edit `ansible-st2` as well.

The better approach is to have `st2chatops_config` hash/dict which will allow us to configure *ANY* env variable in `st2chatops.env`.
Take a look at my comment, we need the following syntax to work: https://github.com/StackStorm/ansible-st2/pull/125#issuecomment-284414479

```sh
# hash to configure any value in st2chatops.env
# https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env
st2chatops_config:
  ST2_API_KEY: 123asd456
  HUBOT_ADAPTER: slack
  HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE
  # add here any other var
```

At a high level we just have `{{ st2chatops_config }}` var to rule all config settings.
Similar approach is implemented in https://github.com/StackStorm/ansible-st2/pull/121/files

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    replace: '\1'
+  when: st2chatops_api_key.changed and st2chatops_config.HUBOT_ADAPTER == "slack"
+  register: uncomment_slack_adapter
+  tags: [st2chatops, skip_ansible_lint]
+
+- name: Uncomment and replace Hubot Slack Token in st2chatops.env
+  become: yes
+  replace:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    regexp: '^# (export HUBOT_SLACK_TOKEN.).*'
```
armab: This is still not scalable and handling every single env var is error-prone.

We need better approach.
When `st2chatops_config` "dict" is provided, you just iterate over key/value pairs and set them in `st2chatops.env`.

armab: So with `{{ st2chatops_config }}` user could midify any existing env variable from the https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env.

This should work ([`#commented` vars](https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env#L93-L95)): 
```yaml
st2chatops_config:
  HUBOT_ADAPTER: irc
  HUBOT_IRC_SERVER: example-server.com
  HUBOT_IRC_PORT: 100500
```

With that, keep in mind to cover corner cases like handling both "`#commented`" and "uncommented" records in `st2chatops.env`. Eg. if user provided `HUBOT_ADAPTER: IRC` in `{{ st2chatops_config }}`, - we should uncomment that var and set value. If record is already uncommented, - just set new val.

armab: @humblearner Another corner case.
If user provided completely new env var like `FOO: bar`, - should we write that in `st2chatops.env` too or maybe throw an error that such env var is not available. WDYT?

(new vars):
```
st2chatops_config:
  FOO: bar
```

roles/st2chatops/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+# st2chatops version to install. Use `latest` to get automatic updates or pin it to numeric version like `2.2.0`
+st2chatops_version: latest
+
+# hash to configure any value in st2chatops.env
+# https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env
+st2chatops_config:
+  ST2_API_KEY: CHANGE-ME-PLEASE
+  HUBOT_ADAPTER: slack
```
armab: I think we shouldn't set `HUBOT_ADAPTER: slack` by default if user didn't provide any custom `key: value` for that.
WDYT?

armab:
And yeah, we want to include `st2chatops` role in existing TravisCI machinery otherwise there is no way to know if things really worked.
(you can set "private" ENV variables in Travis UI for Slack).

roles/st2chatops/defaults/main.yml
lakshmi-kannan: These are supposed to contain sane and safe defaults. I am fine with this TBH.

roles/st2chatops/tasks/main.yml
lakshmi-kannan: I'd rather not go through all of this and have a .j2 file for each adapter in files/ with jinja templates for the settings. So based on the adapter variable you can generate a config from ansible with variables substituted and placed in the st2chatops.env file on the box.

armab: The problem with that approach is described here: https://github.com/StackStorm/ansible-st2/pull/125#discussion_r104658131

Ex: @emedvedev merges Mattermost support (https://github.com/StackStorm/st2chatops/pull/65/files) and we obviously won't include it in `ansible-st2`.

Another example is that the list of vars we list in `st2chatops.env` is not full and can't be. There are a bunch of custom env vars for this or that adapter and even Hubot itself.
Just a quick example from search:
- https://github.com/github/hubot/blob/498d65611388d2881e6ae24f74bd68a832a24211/src/robot.coffee#L413-L417
- https://github.com/github/hubot/blob/498d65611388d2881e6ae24f74bd68a832a24211/bin/hubot#L24-L30

Who knows about more vars hidden in the code?
I don't even want to care about it, it's user's problem. They know what they need, - let's give an instrument.

> BTW this approach is similar to https://github.com/StackStorm/ansible-st2/pull/121 to avoid relying on `st2.conf` templating which can become outdated quickly.

lakshmi-kannan: >I don't even want to care about it, it's user's problem. 

No. We explicitly decided that we'll supply this in commented form in our config file so they don't have to go guess. If someone is adding support for an adapter here, they should include a .j2 file. A .j2 file is a schema for the adapter configuration. You won't have the following problem if you went with this approach.

> should we write that key:value in st2chatops.env too or maybe throw an error that such env var is not available. WDYT?

> Who knows about more vars hidden in the code?

Exactly the problem we should solve and not leave it to user

>we obviously won't include env vars in ansible-st2.

I'd validate the "adapter" variable against a set of supported adapters. A supported adapter means presence of .j2 file.

armab: > Exactly the problem we should solve and not leave it to user

I just demonstrated with links how'd fail to do that.
We can recommend to set some frequently used vars in st2chatops.env but we literally can't inspect every new hubot adapter for some hidden vars. And even by adding new vars you can't monitor all those adapters when new env var is added or new feature is implemented.

Additionally, in st2chatops.env user can add any new env var he wants. Think about custom hubot plugins enabled which is usually the case for real installations. So env vars in reality are used not just to configure predefined list of adaptors, but much more.
By hardcoding we bring more work for us and less flexibility for users.

> If someone is adding support for an adapter here

None will be adding support for adapter here after adding it in st2chatops or hubot-stackstorm. I had enough cases to be realistic about this when different repos become unsync quickly.
That's the reason I'm implementing config setting in a "free form" https://github.com/StackStorm/ansible-st2/pull/121 rather then hardcoded key-vals for every setting. And I would like us to be consistent in this approach across entire Ansible repository to keep the configuration repo management realistic.

armab: Just googled a bit more about other implementations and "what could be expected".

See this config examples, looks pretty reasonable:
Ansible example for Hubot: https://github.com/brianshumate/ansible-hubot
Chef example for Hubot: https://github.com/chef-cookbooks/hubot#attributes

I think we can do it in the same way: 
Make `{{ st2chatops_adapter }}` var required (since hubot won't work without adapter) and additional env vars via `{{ st2chatops_config }}` dict optional.

The problem with hardcoded vars is that apart of "adapter vars" there are also hubot plugin vars:
https://github.com/brianshumate/ansible-hubot/blob/5a1d679472ebfd12119cc9db0209a219560cc6ab/defaults/main.yml#L61

armab:
Just a note:

As a good configuration management citizens we'll need ansible vars to install & enable hubot external scripts/plugins from http://hubot-script-catalog.herokuapp.com/
As said before, that's how real ChatOps deployment usually looks like (ex: `!shipit` from the Hubot community and more stuff like that). That's where custom ENV vars will be needed.

Same as it's implemented in https://github.com/brianshumate/ansible-hubot and https://github.com/chef-cookbooks/hubot#attributes

Not a requirement for this PR, - can wait for future.

humblearner:
Alright!! After long discussion and tie-breaker (Thanks @bigmstone!!). We are going with `dict` and `loop_control`.

For now, I have kept the [Configure adapter setting](https://github.com/StackStorm/ansible-st2/pull/125/files#diff-b9eaf1c4002cf73f79f7506ed1c4f718R52) task in `main.yml`. In case it gets complex we can move it to its own file `config.yml`.

FYI.. going with `.j2` would have been more painful and complex as individual `template` modules for adapters would've overwritten the env file. Using `assemble` module meant more tasks to get all the parameters besides adapter settings in st2chatops.env.

ToDo:

- [ ] Add check for every key in the hash: Currently all the keys should have adapter name `export HUBOT_<Adapter_Name>_<Variable>=<Value>` in `st2chatops.env`.
- [ ] Add TOKEN in TravisUI. The token is hidden in the [Travis log](https://travis-ci.org/StackStorm/ansible-st2/jobs/209604438#L1073)
```
TASK [st2chatops : Add st2_api_key in st2chatops.env] **************************
task path: /vagrant/roles/st2chatops/tasks/main.yml:18
changed: [centos6] => {"changed": true, "msg": "1 replacements made"}
```
and clean: https://github.com/StackStorm/ansible-st2/pull/125/files#diff-ec9a9d2344ff9779bfa7756c4ada61feR5
- [ ] Update README

armab:
> `changed: [localhost] => {"changed": true, "msg": "1 replacements made"}`

@humblearner Yeah, that sounds good enough. I thought it will show real API_KEY in the log.

Vault is not necessarry, but as additional precautionary measure adding [`no_log: yes`](http://docs.ansible.com/ansible/faq.html#how-do-i-keep-secret-data-in-my-playbook) for specific task that saves potentionally sensitive data in `st2chatops.env` would be nice.

armab:
> Add check for every key in the hash: Currently all the keys should have adapter name export HUBOT_<Adapter_Name>_<Variable>=<Value> in st2chatops.env.

I'm not sure if we need such validation, since with `{{ st2chatops_config }}` user potentionally can save any ENV variable he wants in `st2chatops.env`. As said before, custom ENV variables are used not only to configure specific hubot adapters, but also any specific hubot plugins installed by the user, see: http://hubot-script-catalog.herokuapp.com/

There are a bunch of variables you can set, depending on Hubot plugins installed: 
```sh
export FILE_BRAIN_PATH
export HUBOT_SHIP_EXTRA_SQUIRRELS
export HUBOT_ANNOUNCE_ROOMS
```
and so we can't and don't want to control that via any kind of rules.

roles/st2chatops/meta/main.yml
```diff
@@ -0,0 +1,24 @@
...
+        - xenial
+    - name: EL
+      versions:
+        - 6
+        - 7
+  categories:
+    - st2
+    - devops
+    - chatops
+    - autmation
```
armab: automation

roles/st2chatops/meta/main.yml
```diff
@@ -0,0 +1,24 @@
...
+dependencies:
+    - role: st2
```
armab: `st2` is not a requirement for `st2chatops` since `st2chatops` can be installed on a separated machine, it's possible to configure [remote API endpoints](https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env#L22-L26) and it will work.

Saying that, `st2repo` is a dependency for sure in install st2chatops package.

---

cc @lakshmi-kannan you asked before why we need `st2repo` and `st2` roles separated.

roles/st2chatops/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
+# st2chatops version to install. Use `latest` to get automatic updates or pin it to numeric version like `2.2.0`
+st2chatops_version: latest
+
+# Please provide ST2_API_KEY using "st2 apikey create -k"
+st2chatops_ST2_API_KEY: "CHANGE-ME-PLEASEdd"
+
+# Hubot Adapter to be used for st2chatops
+st2chatops_HUBOT_ADAPTER: CHANGE-ME-PLEASE
```
armab: Just an empty string or None (which won't set the val in `st2chatops.env`) as a default would be nice.

Also I'm more in favor of lowercase vars like: `st2chatops_adapter` or if you want `st2chatops_hubot_adapter`.

roles/st2chatops/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
+# st2chatops version to install. Use `latest` to get automatic updates or pin it to numeric version like `2.2.0`
+st2chatops_version: latest
+
+# Please provide ST2_API_KEY using "st2 apikey create -k"
+st2chatops_ST2_API_KEY: "CHANGE-ME-PLEASEdd"
```
armab: Same as previous comment, just an empty string or maybe `None` (which won't set the val in st2chatops.env) instead of `CHANGE-ME-PLEASEdd` will look nicer.

roles/st2chatops/meta/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+galaxy_info:
+  description: Install st2chatops
+  author: humblearner
+  company: StackStorm
+  license: Apache
```
armab: Apache 2.0

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+
+- name: Add st2_api_key in st2chatops.env
+  become: yes
+  replace:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    regexp: '(?<=ST2_API_KEY=)(.*)$'
+    replace: '{{ st2chatops_ST2_API_KEY }}'
+  when: st2chatops_ST2_API_KEY != "CHANGE-ME-PLEASE"
+  register: st2chatops_api_key
+  tags: st2chatops
```
armab: This task should trigger `st2chatops restart` handler since it changes `st2chatops.env` settings.

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+
+- name: Comment Username, Password and Auth URL, if API_KEY provided
```
armab: Good check :+1:

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+  replace:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    regexp: '{{ item }}'
+    replace: '# \1'
+  with_items:
+    - '^(export ST2_AUTH_URL.*)'
+    - '^(export ST2_AUTH_USERNAME.*)'
+    - '^(export ST2_AUTH_PASSWORD.*)'
+  when: st2chatops_api_key.changed
+  tags: [st2chatops, skip_ansible_lint]
```
armab: Same as before, any touch of `st2chatops.env` should trigger `st2chatops restart` handler.

roles/st2chatops/vars/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+supported_hubot_adapter:
```
armab: supported_hubot_adapter*s*

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+
+- name: Uncomment Hubot "{{ st2chatops_HUBOT_ADAPTER|upper }}" adapter in "/opt/stackstorm/chatops/st2chatops.env"
+  become: yes
+  replace:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    regexp: '^# (export HUBOT_ADAPTER={{ st2chatops_HUBOT_ADAPTER }})$'
+    replace: '\1'
+  when: st2chatops_api_key.changed and st2chatops_HUBOT_ADAPTER in supported_hubot_adapter
```
armab: `st2chatops_HUBOT_ADAPTER in supported_hubot_adapter`

In this case it will skip the task if adapater is wrong.
I think it makes sense to move this validation logic in separated task and FAIL playbook execution early and loud if user provided invalid adapter.

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,71 @@
...
+  ini_file:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    section: null
+    option: "export {{ _conf_option.key }}"
+    value: "{{ _conf_option.value }}"
+    no_extra_spaces: yes
+  with_dict: "{{ st2chatops_config }}"
+  loop_control:
+    loop_var: _conf_option
+  when: uncomment_adapter.changed  # TODO: Add check for Individual keys in hash
```
armab: This will fail in cases when user can re-configure adapter settings.
Ex: change API key, but didn't change adapter.

Consider this Ansible role not as a way to install `st2chatops` role only once, but in a wide conext of management where different values could be changed/reconfigured with time.

README.md
```diff
@@ -43,10 +43,13 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2_auth_password`      | `testp`       | Password used by StackStorm standalone authentication.
 | `st2_save_credentials`   | `yes`         | Save credentials for local CLI in `/root/.st2/config` file.
 | **st2mistral**
-| `st2mistral_version`     | `latest`      | st2mistral version to install. Use latest `latest` to get automatic updates or pin it to numeric version like `2.1.1`.
+| `st2mistral_version`     | `latest`      | st2mistral version to install. Use latest `latest` to get automatic updates or pin it to numeric version like `2.2.0`.
 | `st2mistral_db`          | `mistral`     | PostgreSQL DB name for Mistral.
 | `st2mistral_db_username` | `mistral`     | PostgreSQL DB user for Mistral.
 | `st2mistral_db_password` | `StackStorm`  | PostgreSQL DB password for Mistral.
+| **st2chatops**
+| `st2chatops_config`      | `{ }`         | Hash of st2chatops settings to edit in [`st2chatops.env`](https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env) file. For example, `Slack` hubot adapter requires: `ST2_API_KEY: CHANGE-ME-PLEASE` `HUBOT_ADAPTER: slack` `HUBOT_SLACK_TOKEN: xoxb-CHANGE-ME-PLEASE`
```
armab: With new vars, seems `ST2_API_KEY` and `HUBOT_ADAPTER` are irrelevant now.

roles/st2chatops/meta/main.yml
armab: Also i think we can include additional `hubot` tag.

armab:
Curious how `st2chatops` is up and running without providing requirements like ST2_API_KEY and other real settings. Probably the process exits in a couple of seconds later.

We need to add some simple check in `st2smoketests` to keep it real.

roles/st2chatops/vars/main.yml
```diff
@@ -0,0 +1,9 @@
...
+ - slack
+ - hipchat
+ - xmpp
+ - flowdock
+ - yammer
+ - spark
+ - irc
```
armab: Please also include `shell` adapter: https://github.com/github/hubot/blob/master/docs/adapters/shell.md

Do you think it will be better if we use `shell` adapter instead of Slack for testing/smoketests purposes in TravisCI?

armab: Also `mattermost` adapter.
Just to be one step ahead because of https://github.com/StackStorm/st2chatops/pull/65
I'm sure we'll forget to add it later.

armab:
When all finished/merged, we'll need [Ansible `st2docs`](https://github.com/StackStorm/st2docs/blob/master/docs/source/install/ansible.rst) update to highlight `st2chatops` installation and configuration.

roles/st2chatops/defaults/main.yml
humblearner: `CHANGE-ME-PLEASE` is similar to https://github.com/brianshumate/ansible-hubot/blob/master/defaults/main.yml#L58. Better explicit than implicit. 🙂

roles/st2chatops/defaults/main.yml
humblearner: https://github.com/StackStorm/ansible-st2/pull/125/files#r105474907

humblearner:
> @humblearner Another corner case.
> If user provided completely new env var like FOO: bar via {{ st2chatops_config }}, - should we write that key:value in st2chatops.env too or maybe throw an error that such env var is not available. WDYT?
> (new vars):
> ```
> st2chatops_config:
>  FOO: bar```
 is why i had ToDo task: 
> Add check for every key in the hash: Currently all the keys should have adapter name export HUBOT_<Adapter_Name>_<Variable>=<Value> in st2chatops.env.

However, I am perfectly fine, leaving it open for the user to decide, what they want to add to their hubot adapter settings.

armab:
Yeah, I think that corner case is irrelevant now, considering research and fact that env variable could be anything from the http://hubot-script-catalog.herokuapp.com/, not just adapter-specific.

roles/st2chatops/meta/main.yml
humblearner: Apparently, it should but it pukes:

```
RUNNING HANDLER [st2chatops : restart st2chatops] ******************************
fatal: [centos6]: FAILED! => {"changed": false, "failed": true, "msg": "touch: cannot touch `/var/log/st2/st2chatops.log': No such file or directory\nrunuser: group st2 does not exist\n"}
```

armab: Nice! This sounds like a bug discovered.

st2chatops can be configured on separated machine (https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env#L23)  and should not rely on resources created by st2 package.

Can you document the bug in st2chatops repo? It's a packaging issue, we can fix it later.

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,96 @@
...
+  lineinfile:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    line: export HUBOT_ADAPTER=shell
+  when: st2chatops_hubot_adapter == "shell"
+  notify: restart st2chatops
+  register: task_shell_adapter
+  tags: st2chatops
+
+
+- name: Uncomment Hubot "{{ st2chatops_hubot_adapter|upper }}" adapter in "/opt/stackstorm/chatops/st2chatops.env"
```
armab: What if i specify `st2chatops_hubot_adapter=slack` in Ansible and then change it to `st2chatops_hubot_adapter=hipchat` or any other adapter later?

Will it work?

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,96 @@
+---
+- name: Assert st2chatops_hubot_adapter is specified
+  fail:
+    msg: >
+      '"st2chatops_hubot_adapter" must be supplied in roles/st2chatops/defaults/main.yml for st2chatops installation.
```
armab: We can omit mentioning `roles/st2chatops/defaults/main.yml` since it's not single point of truth.
Just asking user to provide `{{ st2chatops_hubot_adapter }}` var is enough.

armab:
Sync-up with `master` to include the recent [build failure fix](https://github.com/StackStorm/ansible-st2/pull/127).

Additionally, as said before, we'll need some smoke test case to make sure `st2chatops` really works.

roles/st2chatops/tasks/main.yml
humblearner: I guess, you mean `https://github.com/StackStorm/ansible-st2/blob/master/roles/st2chatops/vars/main.yml`

roles/st2chatops/vars/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+supported_hubot_adapters:
+ - slack
+ - hipchat
+ - xmpp
+ - flowdock
+ - yammer
+ - spark
+ - irc
+ - mattermost
```
armab: Just discovered via `st2chatops` PR that the right naming for mattermost adapter is `matteruser`, see: https://github.com/StackStorm/st2chatops/blob/1164c0823885931e3a3ad48e0cbb52a336f2d01b/st2chatops.env#L107

emedvedev: FWIW, `mattermost` adapter should work as well, and we have support for it in hubot-stackstorm, but @armab is right, `matteruser` is the one we pack in `st2chatops`.

roles/st2smoketests/tasks/main.yml
```diff
@@ -49,3 +49,104 @@
...
+
+- name: Update st2_api_key in st2chatops.env
+  become: yes
+  replace:
+    dest: /opt/stackstorm/chatops/st2chatops.env
+    regexp: '(?<=ST2_API_KEY=)(.*)$'
+    replace: '{{ st2chatops_api_key }}'
+  no_log: true
+  changed_when: no
+  notify: restart st2chatops
```
humblearner: This is a catch-22 situation. Either restart or idempotency. Removing this and https://github.com/StackStorm/ansible-st2/pull/125/files#diff-c6fa708e733eecbe8e8a283eb5690c19R124 in favor of: https://github.com/StackStorm/ansible-st2/pull/125/files#diff-c6fa708e733eecbe8e8a283eb5690c19R140

roles/st2smoketests/tasks/main.yml
```diff
@@ -49,3 +49,102 @@
...
+  tags:
+    - smoke-tests
+
+- meta: flush_handlers
+  tags:
+    - smoke-tests
+
+- name: Verify st2chatops using bin/hubot
+  shell: "cd /opt/stackstorm/chatops/ && { echo -n; sleep 5; echo '! st2 list actions'; echo; sleep 10; } | bin/hubot --test "
```
armab: With `bin/hubot --test` we actually test `shell` adapter.
See: https://github.com/StackStorm/st2chatops/blob/master/bin/hubot#L20-L22

So decide whether you test `slack` or `shell` hubot adapter.

roles/st2chatops/tasks/main.yml
```diff
@@ -0,0 +1,115 @@
+---
+- name: Assert st2chatops_hubot_adapter is specified
+  fail:
+    msg: >
+      '"st2chatops_hubot_adapter" must be supplied in roles/st2chatops/defaults/main.yml for st2chatops installation.'
```
armab: This error message looks not exactly what we check, according to:
```
when: st2chatops_hubot_adapter not in supported_hubot_adapters
```
conditional.

As I understand we really check if `{{ st2chatops_hubot_adapter }}` is supported, but throw an error saying that `{{ st2chatops_hubot_adapter }}` is not provided in `roles/st2chatops/defaults/main.yml`.

armab:
Once merged what we'll need also:
- `st2chatops` issue
- `st2docs` PR (https://github.com/StackStorm/ansible-st2/pull/125#issuecomment-285691469)

Issue #45, opened 2016-05-06
Split into `st2`, `st2mistral`, `st2web`, `st2chatops`


armab:
Add more flexibility, distribute as different Ansible Galaxy roles (?)

#### Some arguments: 
* StackStorm can be used in [High Availability Deployment](https://docs.stackstorm.com/1.4/reference/ha.html)
* Could be installed/distributed across many hosts
* Can use external RabbitMQ, PostgreSQL, Mongo, ChatOps server
* Not all packages are needed for everyone
* Less complexity if more distros will be added #5

cognifloyd:
How about we do both? One main repo for development and unified use with read-only subsplit repos for each of the roles? We could use something like:
https://github.com/dflydev/git-subsplit

cognifloyd:
I learned about git-subsplit from the Neos community. They use [Neos/slicer](https://github.com/neos/slicer) (which uses git-subsplit) to manage their split repos. I just asked, and they'd like to move to from git-subsplit to [splitsh/lite](https://github.com/splitsh/lite), but the tools create different commit hashes, so they couldn't complete the transition. We should probably start with slitsh/lite to begin with as it is supposed to be much faster.

cognifloyd:
Right now, I'm mixing the stackstorm roles in a custom playbook that intersperses some of my own roles. I installed `StackStorm.stackstorm` from galaxy. To use the stackstorm roles individually, however, I have to refer to them as `StackStorm.stackstorm/roles/<role>`, and if I want to replace one of the stackstorm provided roles, it might still run the stackstorm one, because of meta dependencies. ie: `StackStorm.stackstorm/roles/mongodb` is not `mongodb`. Publishing them as individual roles makes it easier to reuse, or override roles selectively.

Current meta dependencies:
- https://github.com/StackStorm/ansible-st2/blob/9d34eca383d40c63ecfe021f0a593d7d918660f8/roles/mongodb/meta/main.yml#L19-L21
- https://github.com/StackStorm/ansible-st2/blob/01bd5352d2310fd0881fe5cb30b6dd199848c80d/roles/st2/meta/main.yml#L23-L26
- https://github.com/StackStorm/ansible-st2/blob/297e2eb89b585f0bebca6d5f6599d47ed3fcfa2e/roles/st2chatops/meta/main.yml#L23-L25
[...]

PR #126, opened 2017-03-07
Setup BWC repos, Install BWC and Setup RBAC

lakshmi-kannan:
To test, I changed stackstorm.yml to include following:

```
    - name: bwc_repos
      vars:
        license: <REDACTED>

    - name: bwc
      vars:
        ldap: 
          config:
            bind_dn: foo
            ...
        rbac:
          roles:
              - foo
          assignments:
              - bar
```

TODO:

* [X] Rotate license keys after ensuring we don't log anymore
* [x] Idempotency fails in U14 and U16  https://gist.github.com/lakshmi-kannan/581bf90b3a192ae7303b662ad197bc13
* [x] bwc smoke tests


armab:
What do you think about moving `bwc_rbac` role logic inside the `bwc`?

`RBAC` comes tightly connected with `bwc-enterprise` package/dependencies in any way. Looks a bit more simpler: 
```
  roles:
    - name: Install and Configure Brocade Workflow Composer
      role: bwc
      vars:
        bwc_version: latest
        bwc_rbac_enable: yes
        bwc_rbac_roles:
          ...
        bwc_rbac_assignments:
          ...
```

which is similar to `st2` role with [`st2_auth_enable: yes`](https://github.com/StackStorm/ansible-st2#variables`) var.

Related question to that. Do you think we'll have more roles related to `bwc` in nearest future? Just to have better understanding, since it could be a game changer how to distribute everything later (you previous valid concern about: roles/galaxy/repos).

roles/bwc_rbac/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+
+st2_config_file_path: /etc/st2/st2.conf
```
armab: I think we can hardcode it without any big worries. `st2.conf` path is pretty much a constant everywhere, bundled in packaging.

armab:
Just a note: We should think about better approaches for setting `RBAC` configs in this or maybe in future `bwc` role versions.

Vars like: 
```
bwc_superuser: st2admin
rbac_setup_default_assignments: true
```
seems to provide not that much flexibility in terms of configuration. But yeah, it's a good start.

armab:
@lakshmi-kannan Returning to your original concern about roles/galaxy/repos/distribution. Here is a more detailed write-up about our options:

### 1) `BWC` role in separated repo
*Long-term (?), more effort, polishing, time*

- As you noted, we can use [packagecloud-ansible-role](https://github.com/computology/packagecloud-ansible-role) as external dependency for `bwc`
  - but it's not idempotent
  - in future we need contribute PRs to improve it
- Publish bwc on Ansible-Galaxy later
   - important requirements (!): `bwc` should work as a single role (with depenencies if needed), very different approach comparing to `ansible-st2`
- Blog Post it later (marketing ++)
- Harder to test the role, since we need st2 installed first
- Needs some CI configuration
- Long term solution, harder to do/more to polish/more time

### 2) `BWC` in `ansible-st2` repo
*KISS, faster, easier for short-term*

- I don't like [packagecloud-ansible-role](https://github.com/computology/packagecloud-ansible-role) code and how we re-use it in `ansible-st2`, needs much more polishing.
- Easy to test, since all roles are in one repo and CI is configured: https://travis-ci.org/StackStorm/ansible-st2
- short-term solution, easier to do if we need it faster

I'll let you to implement/polish the main "core" logic (since it's needed in any way) and later when it's finished I think we should throw all pros/cons to decide what works best based on improved data/arguments/ideas/priorities we have.
The reasoning about rising these questions, - I have opinions and my requests to change this or that piece of code depends on the approach we choose.

> PS. Take a look at related: https://github.com/StackStorm/ansible-st2/issues/45 to have better context.

roles/bwc/tasks/ldap.yml
```diff
@@ -0,0 +1,22 @@
...
+    value: ldap
+    backup: yes
+
+- name: Setup st2.conf auth backend_kwargs for LDAP
+  become: yes
+  ini_file:
+    dest: "{{ st2_config_file_path }}"
+    section: auth
+    option: backend_kwargs
+    value: "{{ ldap.config | to_nice_json }}"
```
lakshmi-kannan: This prints out a multi-line string in the ini file and our config parser pukes. I tried to_json. Still the same.

roles/bwc/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+bwc_pkg_repo: "enterprise"
+# 'latest' to get latest version or numeric like '2.1.1'
+bwc_version: latest
+# used only if 'bwc_version' is numeric
+bwc_revision: 1
+
+st2_config_file_path: /etc/st2/st2.conf
```
armab: As noted before, don't forget to hardcode `st2.conf` path.
We don't need it in vars (https://github.com/StackStorm/ansible-st2/pull/126#discussion_r104986082 reasons why).

roles/bwc/tasks/ldap.yml
```diff
@@ -0,0 +1,22 @@
+---
+
+- name: Setup st2.conf auth backend to LDAP
+  become: yes
+  ini_file:
+    dest: "{{ st2_config_file_path }}"
+    section: auth
+    option: backend
+    value: ldap
+    backup: yes
```
armab: Notify st2 `restart` or `reload` handler (if enough) trigger when changing the `st2.conf` file.
Example: https://github.com/StackStorm/ansible-st2/blob/master/roles/st2/tasks/auth.yml#L34

roles/bwc/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  become: yes
+  package:
+    name: bwc-enterprise={{ bwc_version }}-{{ bwc_revision }}
+    state: present
+  when: bwc_version != "latest"
+  tags:
+    - bwc
+    - st2 enterprise
+
+- name: Setup RBAC and setup roles and assignments if enable_rbac is defined
```
armab: What is `enable_rbac`?

roles/bwc/tasks/rbac.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  - name: Copy RBAC roles to /opt/stackstorm/rbac/roles directory
+    become: yes
+    template:
+      src: rbac_roles/roles.yml.j2
+      dest: /opt/stackstorm/rbac/roles/{{ item.name }}.yaml
+      owner: st2
+      group: st2
+    with_items: "{{ rbac.roles }}"
+    when: rbac_roles is defined
```
armab: Can't find any occurrence of `rbac_roles` var. Is it used anywhere?

roles/bwc/tasks/rbac.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  - name: Copy RBAC assignments to /opt/stackstorm/rbac/assignments directory
+    become: yes
+    template:
+      src: rbac_assignments/assignments.yml.j2
+      dest: /opt/stackstorm/rbac/assignments/{{ item.name }}.yaml
+      owner: st2
+      group: st2
+    with_items: "{{ rbac.assignments }}"
+    when: rbac_assignments is defined
```
armab: Can't find any occurrence of `rbac_assignments` var. Is it used anywhere?

roles/bwc/tasks/rbac.yml
```diff
@@ -0,0 +1,47 @@
...
+
+  - name: Enable RBAC in st2 configuration
```
armab: Enable auth maybe?

roles/bwc_repos/tasks/bwc_repos_apt.yml
```diff
@@ -0,0 +1,40 @@
...
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - debian-archive-keyring
+    - apt-transport-https
+
+- name: Get GPG key URL for private repo
+  become: yes
+  shell: 'curl "{{ deb_gpg_key_url }}"'
```
armab: This is not idempotent

roles/bwc_repos/tasks/bwc_repos_apt.yml
```diff
@@ -0,0 +1,40 @@
...
+  register: deb_gpg_key
+
+# This is the exact key as the open source repo but this behavior might change. So just
+# following what's in packagecloud docs for private repos.
```
armab: Right, and it's not idempotent.

Please refactor `bwcrepos` role following the `st2repos` logic where possible.

Ex for this case: https://github.com/StackStorm/ansible-st2/blob/91bd7189500c36377e50b3cfaf20c4dbe6ef1f58/roles/st2repos/tasks/st2repos_apt.yml#L11-L16

roles/bwc_repos/tasks/bwc_repos_apt.yml
```diff
@@ -0,0 +1,40 @@
+---
+
+- name: Assert that master_token is specified
+  fail: msg="License key must be supplied for BWC enterprise installation."
+  when: license is not defined
```
armab: Please include all possible vars used in the roles in `defaults.yml` with comments and decent README.
With that, instead of relying on `is not defined` we can rely on is empty conditional check.

armab: Additionally, vars should follow the `{{ role_name}}_var_name` naming.
I guess for `bwc` role ideal var could be `{{ bwc_license }}`. And yeah, I must admit that for `bwc_repo` or `bwc_repos` role var name `{{ bwc_repo_license }}` looks a bit ugly :/

roles/bwc_repos/tasks/bwc_repos_apt.yml
```diff
@@ -0,0 +1,40 @@
...
+- name: Get GPG key URL for private repo
+  become: yes
+  shell: 'curl "{{ deb_gpg_key_url }}"'
+  register: deb_gpg_key
+
+# This is the exact key as the open source repo but this behavior might change. So just
+# following what's in packagecloud docs for private repos.
+- name: Add GPG key to keyring
+  become: yes
+  shell: 'curl -L "{{ deb_gpg_key.stdout }}" | apt-key add -'
```
armab: not idempotent

roles/bwc_repos/tasks/bwc_repos_apt.yml
```diff
@@ -0,0 +1,40 @@
...
+  when: ansible_os_family == "Debian" and license is defined
```
armab: We never reach this task if license is not defined, because of `fail` task in this playbook.

armab: > `ansible_os_family == "Debian"`

We're already in Debian play, according to file name `bwc_repos_apt.yml`.

roles/bwc_repos/tasks/bwc_repos_yum.yml
```diff
@@ -0,0 +1,43 @@
+---
+
+- name: Assert that master_token is specified
+  fail: msg="License key must be supplied for BWC enterprise installation."
```
armab: `fail` check is duplicated in both `bwc_repos_yum.yml` and `bwc_repos_apt.yml` files.
Sounds cool if we can juse move it to parent `main.yml` file.

roles/bwc_repos/tasks/bwc_repos_yum.yml
```diff
@@ -0,0 +1,43 @@
...
+  become: yes
+  shell: 'curl "{{ rpm_config_file_url }}" > {{ rpm_config_file_location }}'
+  args:
+    creates: "{{ rpm_config_file_location }}"
+  register: added_bwc_rpm_repository
+  when: ansible_os_family == "RedHat" and license is defined
+
+- name: Update yum package cache
+  become: yes
+  shell: yum -q makecache -y --disablerepo='*' --enablerepo='StackStorm_{{ bwc_pkg_repo|replace("/", "_") }}'
```
armab: Any idea why do we need this?
Also it's not idempotent.

roles/bwc_repos/tasks/bwc_repos_yum.yml
```diff
@@ -0,0 +1,43 @@
...
+  when: ansible_os_family == "RedHat" and added_bwc_rpm_repository|success
+
+- name: Update yum package cache BWC enterprise source repo
+  become: yes
+  shell: yum -q makecache -y --disablerepo='*' --enablerepo='StackStorm_{{ bwc_pkg_repo|replace("/", "_") }}-source'
+  when: ansible_os_family == "RedHat" and added_bwc_rpm_repository|success
```
armab: > `when: ansible_os_family == "RedHat"`

we're already in os-family specific `yum` task.

roles/bwc/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  include: "rbac.yml"
+  when: rbac is defined
```
armab: Please name vars accordingly to role, eg: `bwc_rbac`.

roles/bwc/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+- name: Setup LDAP and set up LDAP configuration
+  include: "ldap.yml"
+  when: ldap is defined
```
armab: Please name vars accordingly to role, eg: `bwc_ldap`.

roles/bwc/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+bwc_pkg_repo: "enterprise"
+# 'latest' to get latest version or numeric like '2.1.1'
+bwc_version: latest
+# used only if 'bwc_version' is numeric
+bwc_revision: 1
+
```
armab: Please define all vars used in this role with comments, as well as include them in `README.md`.

armab:
When all finished/merged, we'll need [Ansible `st2docs`](https://github.com/StackStorm/st2docs/blob/master/docs/source/install/ansible.rst) update with separated section to install/configure `bwc`.

roles/bwc/tasks/ldap.yml
lakshmi-kannan: Yeah, we also have to set the backend_kwargs before restarting. See next step. I did notify.

armab: Include notify for every task where its needed logically (ex: changing the conf file), its less error prone since "next task" could be removed, adjusted or never ran because of some conditional.

It doesn't matter how many times we call Ansible handlers, - they will be executed in the end of entire play run in any way.

See "Handlers: Running Operations On Change": http://docs.ansible.com/ansible/playbooks_intro.html with better explanation how it works.

lakshmi-kannan: Ahh, good to know about that.

armab:
Just a note: sync-up with `master` to include the recent [build failure fix](https://github.com/StackStorm/ansible-st2/pull/127).

roles/bwc/tasks/bwc_repos_yum.yml
```diff
@@ -0,0 +1,37 @@
...
+# Fixes "Failure talking to yum: Cannot retrieve repository metadata (repomd.xml) for repository: StackStorm_stable. Please verify its path and try again" when installing st2
+- name: Update ca-certificates package
+  become: yes
+  yum:
+    name: ca-certificates
+    state: latest
+  tags: skip_ansible_lint
+
+# See: https://github.com/docker-library/docs/tree/master/centos#package-documentation
+# We ship `nginx.conf` via `st2` package doc files, for example
```
armab: This block could be removed, was really critical for `st2` package.

roles/bwc/templates/rbac_roles/roles.yml.j2
```diff
@@ -0,0 +1,6 @@
+---
+
+name: {{ item.name }}
+description: {{ item.description }}
+permission_grants:
+    {{ item.permission_grants | to_nice_json }}
```
lakshmi-kannan: See https://gist.github.com/lakshmi-kannan/0f5f073311a40d209d3e19c7cab3e864 for why I had to use to_nice_json.

roles/bwc/tasks/bwc_repos_apt.yml
```diff
@@ -0,0 +1,32 @@
+---
+
+- name: Assert that master_token is specified
+  fail: msg="License key must be supplied for BWC enterprise installation."
```
armab: Not needed in this file, since you include that check in previous step `bwc_repos_setup.yml`

roles/bwc/tasks/bwc_repos_setup.yml
```diff
@@ -0,0 +1,42 @@
...
+    owner: st2
+    group: st2
+    state: directory
+
+- name: Get read token for repo from packagecloud
+  become: yes
+  no_log: yes
+  uri:
+    url: https://{{ bwc_license }}:@packagecloud.io/install/repositories/StackStorm/{{ bwc_pkg_repo }}/tokens.text
+    creates: "/etc/packagecloud/StackStorm_{{ bwc_pkg_repo }}_read_token.txt"  # Don't download if file already exists
```
armab: Following our previous case when we rotated `{{ bwc_license }}`.

What will happen if user changes `{{ bwc_license }}`? Seems like `read token` will remain from the old license.

Eg. if user changes the license - Ansible should update apt/yum repository too with new read token.

roles/bwc/tasks/ldap.yml
```diff
@@ -0,0 +1,32 @@
...
+  # Unfortunately, ``with_dict`` also logs the dict which could leak passwords.
+  no_log: yes
+  ini_file:
+    dest: /etc/st2/st2.conf
+    section: auth
+    option: backend
+    value: ldap
+    backup: yes
+  # Don't even setup LDAP if backend_kwargs is not defined
+  with_dict: "{{ bwc_ldap.backend_kwargs | default({}) }}"
```
armab: I see `with_dict` here but don't see `{{ item }}`. That thing is used only for loops, see http://docs.ansible.com/ansible/playbooks_loops.html#looping-over-hashes.

So you probably want something like: 
```
when: bwc_ldap.backend_kwargs is defined
```

roles/bwc/tasks/ldap.yml
```diff
@@ -0,0 +1,32 @@
...
+  become: yes
+  # Unfortunately, ``with_dict`` also logs the dict which could leak passwords.
+  no_log: yes
+  ini_file:
+    dest: /etc/st2/st2.conf
+    section: auth
+    option: backend_kwargs
+    value: "{{ bwc_ldap.backend_kwargs | to_json | string }}"
+    backup: yes
+  with_dict: "{{ bwc_ldap.backend_kwargs | default({}) }}"
```
armab: Same here, `with_dict` is used only for loops, just: 
```sh
when: bwc_ldap.backend_kwargs is defined
```

roles/bwc/tasks/ldap.yml
armab: Also @lakshmi-kannan `length` for `dict` worked in yaml:

```
TASK [bwc : debug] *************************************************************
task path: /vagrant/roles/bwc/tasks/ldap.yml:6
ok: [ubuntu16] => {
    "{}|length": "0"
}
```

```
TASK [bwc : debug] *************************************************************
task path: /vagrant/roles/bwc/tasks/ldap.yml:6
ok: [ubuntu16] => {
    "{'key':'val','a':'b'}|length": "2"
}
```

roles/bwc/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+    - BWC repos
+    - StackStorm enterprise
+
+- name: Install latest bwc-enterprise package
+  become: yes
+  package:
+    name: bwc-enterprise
+    state: latest
+  register: bwc_installed
+  when: bwc_repo_added|success and bwc_version == "latest"
```
armab: Don't need `bwc_repo_added|success`, since ansible will stop on first failure

roles/bwc/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+    state: present
+  register: bwc_installed
+  when: bwc_repo_added|success and bwc_version != "latest"
+  tags:
+    - bwc
+    - st2 enterprise
+
+- name: Setup RBAC and setup roles and assignments if bwc_rbac is defined
+  include: "rbac.yml"
+  when: bwc_installed|success and  bwc_rbac is defined
```
armab: Don't need `bwc_installed|success` since Ansible will stop on first failure

roles/bwc/tasks/main.yml
```diff
@@ -0,0 +1,38 @@
...
+
+- name: Setup LDAP and set up LDAP configuration
+  include: "ldap.yml"
+  when: bwc_installed|success and bwc_ldap is defined
```
armab: Same here to avoid unecessary check `bwc_installed|success`

roles/bwc/tasks/rbac.yml
```diff
@@ -0,0 +1,64 @@
...
+      state: directory
+    with_items:
+      - /opt/stackstorm/rbac/assignments
+      - /opt/stackstorm/rbac/roles
+
+  - name: Copy default RBAC roles to /opt/stackstorm/rbac/roles directory
+    become: yes
+    template:
+      src: rbac_roles/roles.yml.j2
+      dest: /opt/stackstorm/rbac/roles/{{ item.name }}.yaml
```
armab: Changing any file in `/rbac/` dir should notify `reload bwc_rbac` trigger.

Make sure we run that thing for other tasks in this file.

roles/bwc/templates/rbac_assignments/assignments.yml.j2
```diff
@@ -0,0 +1,5 @@
+---
+
+  username: {{ item.name }}
+  roles:
+    {{ item.roles | to_nice_yaml }}
```
armab: Same as noted in https://github.com/StackStorm/ansible-st2/pull/126/files#r106033276 this should be more correct way:
`{{ item.roles | to_nice_yaml(2) | indent(2) }}`

in case when `item.roles` contains more complex data.

roles/bwc/tasks/bwc_repos_setup.yml
lakshmi-kannan: Yep. Unfortunately, if I generate a token every time, it would cause idempotency failures. If I do what I did, then installation with fail if they swapped the license. But I think this is fine. It fails out with an error. 

Another approach is to ask user for ``{{ bwc_read_token }}`` which I am -1 to.

lakshmi-kannan: Looks like I could force ``changed_when: no``. That sounds hacky to me.

armab: What we can do is to additionally generate the hash from the `{{ bwc_license }}` via http://docs.ansible.com/ansible/playbooks_filters.html#hashing-filters and store hash somewhere. If hash changed, - then we request new `{{ bwc_read_token }}` and save it in repo.

lakshmi-kannan: I think we can address this in a different PR. I am not too keen on adding more scope to this PR.

armab: Deal. Feel free to create an issue for this then so maybe someone else could pick it up.

roles/bwc_smoketests/templates/rbac_assignments/assignments.yml.j2
```diff
@@ -0,0 +1,5 @@
+---
+
+  username: {{ item.name }}
+  roles:
+    {{ item.roles | to_nice_yaml(2) | indent(2) }}
```
armab: indentation should be 2 spaces here

roles/bwc_smoketests/templates/rbac_assignments/assignments.yml.j2
```diff
@@ -0,0 +1,5 @@
+---
+
+  username: {{ item.name }}
+  roles:
```
armab: and 0 spaces here.

Otherwise `{{ item.roles | to_nice_yaml(2) | indent(2) }}` won't work correctly if we add some complex data there

armab:
Thanks for being serious about the quality! :100: 

After merging, what we'll need also is respective `st2docs` change to highlight `bwc` installation (as said in: https://github.com/StackStorm/ansible-st2/pull/126#issuecomment-285691081)

PR #129, opened 2017-03-21
st2_api_key  generation now part of "st2chatops"role and "st2 pack install st2" in smoketests/main.yml

humblearner:
- [x] Move `st2 pack install st2` in `smoketests/main.yaml` as a test, from `st2chatops_smoketests.yml`

- [x] Make API key generation robust. 

Sample st2chatops role in a playbook:
```
- name: Install StackStorm with all services on a single node
  hosts: localhost
  roles:
    - (......other roles......)
    - name: Install st2chatops with "slack" hubot adapter
      role: st2chatops
      vars:
        st2chatops_hubot_adapter: slack
        st2chatops_config: {"HUBOT_SLACK_TOKEN":"xoxb-CHANGE-ME-PLEASE"}
        # optional
        st2chatops_version: latest
        st2chatops_st2_api_key: CHANGE-ME-PLEASE 
        # This can be generated using "st2 apikey create -k" in a pre_task or updated in defaults/main.yml.
        # If not provided will it will be generated in the role, if not already updated.
```

### DIFF
--- a/roles/st2chatops/tasks/main.yml
+++ b/roles/st2chatops/tasks/main.yml
@@ -5,11 +5,13 @@
       '"st2chatops_hubot_adapter" must be one of the supported adapters in
       https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env'
   when: st2chatops_hubot_adapter not in supported_hubot_adapters
+  tags: st2chatops
 
 - name: Assert st2chatops_hubot_adapter "{{ st2chatops_hubot_adapter|upper }}" settings are specified
   fail:
     msg: '"st2chatops_config" hash cannot be empty for "{{ st2chatops_hubot_adapter|upper }}" hubot adapter.'
   when: st2chatops_config.keys() == [] and st2chatops_hubot_adapter != "shell"
+  tags: st2chatops
 
 - name: Install latest st2chatops package
   become: yes
@@ -29,10 +31,11 @@
 
 - name: Check if api key already exist. No change if it already exists
   become: yes
-  shell: "grep -e '^export ST2_API_KEY=\"\\${ST2_API_KEY}\"$' /opt/stackstorm/chatops/st2chatops.env"
+  command: "grep -e '^export ST2_API_KEY=\"\\${ST2_API_KEY}\"$' /opt/stackstorm/chatops/st2chatops.env"
   changed_when: no
   ignore_errors: True
   register: task_apikey_exists
+  tags: st2chatops
 
 - name: Add user defined "st2chatops_st2_api_key" in st2chatops.env, from "role" or "defaults/main.yml"
   become: yes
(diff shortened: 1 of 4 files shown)


roles/st2chatops/tasks/main.yml
```diff
@@ -26,18 +27,55 @@
   when: st2chatops_version != "latest"
   tags: [st2chatops, skip_ansible_lint]
 
-- name: Add st2_api_key in st2chatops.env
+- name: Check if api key already exist. No change if it already exists
+  become: yes
+  shell: "cat /opt/stackstorm/chatops/st2chatops.env | grep -e '^export ST2_API_KEY=\"\\${ST2_API_KEY}\"$'"
```
armab: This looks a bit overcomplicated.

You can omit `cat`, just:
```sh
command: grep -e "expression" /opt/stackstorm/chatops/st2chatops.env
```

Also I see you're using `${ST2_API_KEY}` env variable. Is it defined/populated somewhere?

humblearner: @arma: `${ST2_API_KEY}` from here: https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env#L29 to take care of: `Don't overwrite st2chatops.env if there is existing api key`.

roles/st2chatops/tasks/main.yml
```diff
@@ -26,18 +27,55 @@
...
+  environment:
+    ST2_AUTH_TOKEN: "{{ task_st2_token.stdout }}"
+  when: task_st2_token.changed
+  register: task_generated_api_key
+  no_log: true
+  tags: [st2chatops, skip_ansible_lint]
+
+- name: Add generated "st2_api_key" in st2chatops.env
+  become: yes
+  replace:
```
armab: We're repeating this st2chatops.env API key setup in 2 blocks. Do you think it's possible to improve the logic and include "set API key" task only once?

humblearner: @armab: AFAIU, first "set API key" is for `Add user defined "st2chatops_st2_api_key" in st2chatops.env, from "role" or "defaults/main.yml"` and second for `Add generated "st2_api_key" in st2chatops.env` with different input and conditions. Also, we don't want to generate api key every time we execute this task. The only way that i can think of is moving this task in a separate yaml file.

roles/st2chatops/tasks/main.yml
```diff
@@ -26,18 +27,55 @@
...
+    (st2chatops_st2_api_key is defined) and
+    (st2chatops_st2_api_key != "CHANGE-ME-PLEASE")
+  register: task_user_st2_api_key
   no_log: true
   notify: restart st2chatops
   tags: st2chatops
 
+- name: Generate authentication token
+  command: st2 auth {{ st2_auth_username }} -p {{ st2_auth_password }} -t
+  when: (task_apikey_exists.failed is not defined) and (task_user_st2_api_key.changed == false)
```
armab: And when {{ st2_auth_username }} with {{ st2_auth_password }} both defined.

Otherwise it means that st2chatops role included without st2 role.

And so if there is no way to configure st2chatops, - we should trigger an error with a request to provide st2chatops_st2_api_key var

humblearner: > Otherwise it means that st2chatops role included without st2 role.

We may not hit this until someone deliberately removes `st2` role from `st2chatops` dependencies (even if it is not included in `stackstorm.yml` playbook).. We don't take responsibility if a user edits the meta/tasks unnecessarily. Right now if you provide api_key with the role it will take precedence over generate api task and last one in the line is default.

In case we plan to remove `st2` role as a dependency, we are not handling it and we know where to fix.

armab: You're right :+1: 
I forgot that `st2` role is hard dependency for `st2chatops` until we fix the https://github.com/StackStorm/st2chatops/issues/67

humblearner:
CI fixed here: https://github.com/StackStorm/st2/pull/3309

PR #133, opened 2017-04-14
Add an option to configure custom SSL certificate for st2web/nginx

### LINKED ISSUE #132: Optionally configure custom certificate, instead of generating self-signed [st2web]
Adjust `st2web` role to allow optionally specifying custom cert, instead of generating self-signed one every time.

armab:
Closes #132 

Add 2 new vars for `st2web` role: `st2web_ssl_certificate` and `st2web_ssl_certificate_key`.
If both vars are provided, - save the cert. If not provided, - generate self-signed certificate as it was before.

The default behavior remains as before: generting self-signed cert for `st2web` UI,

### Example usage
```yaml
- name: Install StackStorm
  hosts: all
  roles:
    ...
    - name: Install st2web
      role: st2web
      vars:
        st2web_ssl_certificate: "{{ lookup('file', 'local/path/to/domain-name.crt') }}"
        st2web_ssl_certificate_key: "{{ lookup('file', 'local/path/to/domain-name.key') }}"
```

### TODO
- [ ] Verify for real with a custom cert

### DIFF
--- a/roles/st2web/tasks/certificate.yml
+++ b/roles/st2web/tasks/certificate.yml
@@ -1,0 +1,37 @@
+---
+- name: Verify if custom SSL certificate was correctly specified
+  fail:
+    msg: "When using custom certificate, both 'st2web_ssl_certificate' and 'st2web_ssl_certificate_key' must be provided"
+  # no XOR in Yaml
+  when: (st2web_ssl_certificate and not st2web_ssl_certificate_key) or (not st2web_ssl_certificate and st2web_ssl_certificate_key)
+
+- name: Create SSL certificate directory
+  become: yes
+  file:
+    state: directory
+    dest: /etc/ssl/st2
+    mode: 0700
+
+- name: Save custom SSL certificate
+  become: yes
+  copy:
+    content: "{{ item.cert }}"
+    dest: "{{ item.path }}"
+    mode: 0600
+  with_items:
+    - cert: "{{ st2web_ssl_certificate }}"
+      path: /etc/ssl/st2/st2.crt
+    - cert: "{{ st2web_ssl_certificate_key }}"
+      path: /etc/ssl/st2/st2.key
+  notify:
+    - restart nginx
+  when: st2web_ssl_certificate and st2web_ssl_certificate_key
+
+- name: Generate self-signed SSL certificate
+  become: yes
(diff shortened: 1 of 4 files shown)


roles/st2web/tasks/certificate.yml
```diff
@@ -0,0 +1,37 @@
...
+  when: (st2web_ssl_certificate and not st2web_ssl_certificate_key) or (not st2web_ssl_certificate and st2web_ssl_certificate_key)
+
+- name: Create SSL certificate directory
+  become: yes
+  file:
+    state: directory
+    dest: /etc/ssl/st2
+    mode: 0700
+
+- name: Save custom SSL certificate
```
lakshmi-kannan: @armab This would all be simple if ansible-st2 would allow galaxy imports so we could just use jdauphant's SSL role.

roles/st2web/tasks/certificate.yml
```diff
@@ -0,0 +1,37 @@
...
+  become: yes
+  copy:
+    content: "{{ item.cert }}"
+    dest: "{{ item.path }}"
+    mode: 0600
```
lakshmi-kannan: I'd explicitly set the owner and group just to be clear.

roles/st2web/tasks/certificate.yml
```diff
@@ -0,0 +1,37 @@
...
+      path: /etc/ssl/st2/st2.crt
+    - cert: "{{ st2web_ssl_certificate_key }}"
+      path: /etc/ssl/st2/st2.key
+  notify:
+    - restart nginx
+  when: st2web_ssl_certificate and st2web_ssl_certificate_key
+
+- name: Generate self-signed SSL certificate
+  become: yes
+  shell: openssl req -x509 -newkey rsa:2048 -keyout /etc/ssl/st2/st2.key -out /etc/ssl/st2/st2.crt -days 365 -nodes -subj "/C=US/ST=California/L=Palo Alto/O=StackStorm/OU=Information Technology/CN=$(hostname)"
```
lakshmi-kannan: Wouldn't this generate a new cert every time ansible is run? You may want to stat the file and check if the .crt and .key exist before regeneration.

armab: It has `creates` arg below, so should be good.

roles/st2web/tasks/certificate.yml
armab: Makes sense.
Now wondering what user:group should it be.

Hmm, `root:root` or something different, example: https://superuser.com/a/556496

armab: Changed in https://github.com/StackStorm/ansible-st2/pull/133/commits/e442ebfd4427396ca927a4ff7575a430a3a535c6

PR #140, opened 2017-04-21
Refactor st2chatops smoketests

humblearner:


armab:
There is an alternative approach https://github.com/StackStorm/ansible-st2/pull/143/files#diff-d3fbd21bc6be7ce4531c27a82b2619deR21

We pass needed Ansible vars like `st2chatops_config` and `st2chatops_hubot_adapter` from the Kitchen/CI, so you don't need to edit `st2chatops.env` within st2smoketests.

PR #143, opened 2017-04-27
Refactor Smoketests

armab:
With recent new roles we've introduced some limitations related to how CI/testing can be executed.
This PR is here to experiment with different environments and hopefully improve the testing env, reverting some options for contributors.

- [x] Fix CI for forked PRs (closes #141)
  - [x] If `bwc_license` is not set, - skip the role in Travis
  - [x] Use personal & unencrypted hubot slack token in `travis.yml` (no risk of exposing it to public)
- [x] Cleanup `st2chatops` smoketests (closes #140)
- [x] Users should be able to run `kitchen test` in their local environments
- [ ] ~Smoketests should work with non-default, non-Travis configurations~ (future PRs)
- [ ] Verify if forked PR really works (post-merge)

Issue #141, opened 2017-04-21
CI fails on forked PRs


armab:
TravisCI is red for forked PRs, starting from `st2chatops` and `bwc` role implementations.

This is caused by private env variables defined in our Travis settings: `HUBOT_SLACK_TOKEN` and `BWC_LICENSE` which are not available to forked builds.

https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings
> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

Find the way how to proceed this the right way in future, -  because we can't turn down CI for contribution PRs.

armab:
Also similar: https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions

armab:
Disabled ansible-st2 `bwc` role run for forked PRs :disappointed:

PR #144, opened 2017-05-01
Assert that forked PR CI really works

armab:
Part of the https://github.com/StackStorm/ansible-st2/pull/141 and https://github.com/StackStorm/ansible-st2/pull/143

Check that forked PR CI in Travis really works.
- `st2chatops` role + smoketests should work in forked PR CI
- `bwc` role + smoketests should be skipped for the forked PRs due to license


armab:
Glad I tested that end-to-end after the #143.
Found + fixed more bugs.

PR #148, opened 2017-05-26
Adding present option for version

### LINKED ISSUE #111: Add logic for 'latest' st2_revision
When numeric `st2_version` is used, `st2_revision` is pinned to default val `1`, which is not desired beavior.

Instead, when `st2_revision` is not set: is empty or `latest`, - let playbook install latest package revision number for respective version.

bigmstone:
Name says it all.

Fixes #111

### DIFF
--- a/roles/st2/tasks/main.yml
+++ b/roles/st2/tasks/main.yml
@@ -32,7 +32,7 @@
 - name: Install pinned st2 package
   become: yes
   package:
-    name: st2={{ st2_version }}-{{ st2_revision }}
+    name: st2{{ '-' if ansible_pkg_mgr == 'yum' else '=' }}{{ st2_version }}-{{ st2_revision }}
     state: present
   when:
     - st2_version != "latest"
--- a/roles/st2/tasks/main.yml
+++ b/roles/st2/tasks/main.yml
@@ -18,12 +18,25 @@
     - reload st2
   tags: st2, skip_ansible_lint
 
+- name: Install present st2 package
+  become: yes
+  package:
+    name: st2
+    state: present
+  when: st2_version == "present"
+  notify:
+    - restart st2
+    - reload st2
+  tags: st2, skip_ansible_lint
+
 - name: Install pinned st2 package
   become: yes
   package:
     name: st2={{ st2_version }}-{{ st2_revision }}
     state: present
-  when: st2_version != "latest"
+  when:
+    - st2_version != "latest"
+    - st2_version != "present"
   notify:
     - restart st2
     - reload st2
--- a/roles/st2/tasks/main.yml
+++ b/roles/st2/tasks/main.yml
@@ -7,7 +7,7 @@
   when: ansible_os_family == "RedHat" and ansible_distribution_major_version == "6"
   tags: st2
 
-- name: Install latest st2 package
+- name: Install latest st2 package, auto-update
   become: yes
   package:
     name: st2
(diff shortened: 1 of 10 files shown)


m4dcoder:
LGTM but i should let @humblearner or @armab approve since they're more familiar with this.

roles/st2/tasks/main.yml
```diff
@@ -18,12 +18,25 @@
...
+  notify:
+    - restart st2
+    - reload st2
+  tags: st2, skip_ansible_lint
+
 - name: Install pinned st2 package
   become: yes
   package:
     name: st2={{ st2_version }}-{{ st2_revision }}
     state: present
```
armab: We already have `state: present` logic here.

Thinking about how to use that, instead of introducing another `st2_version: present` keyword.

armab:
After playing a bit with `sudo apt install st2=*` `sudo apt install st2=2.3dev-*` it looks like `*` works for auto-updating, despite Ansible `state: present` flag.

Le me dig in further, including how Ansible `package` module works to make sure the syntax we'll use will be consistent enough across use cases for `st2_version` & `st2_revision`.

bigmstone:
@armab thoughts?
https://github.com/StackStorm/ansible-st2/pull/148/files#diff-555e17548189fd960a2fbb76681e59d9R13

armab:
@bigmstone We've talked about the `st2_version: null` as var value, not `st2_version: present`.

So with `st2_version: null` we do `apt/yum install st2`.

roles/st2/tasks/main.yml
```diff
@@ -7,28 +7,16 @@
   when: ansible_os_family == "RedHat" and ansible_distribution_major_version == "6"
   tags: st2
 
-- name: Install latest st2 package
+- name: Install st2 package
   become: yes
   package:
-    name: st2
-    state: latest
-  when: st2_version == "latest"
```
armab: I'm more about leaving `latest` block as is for less `jinja` hackery.

armab:
Looks there is no easy fix for https://github.com/StackStorm/ansible-st2/issues/111

For example this construction will pin the version, but always try to update the package once there is a new revision available in a repo. So no real pinning happens here: 
```
st2_version: 2.2.1
st2_revision: *
```

Another idea is that we could check if package is installed for `apt`/`yum` separately, parse the version, extract from revision and compare with the user input `st2_version` & `st2_revision`. But this logic easily breaks by the fact that user is free to specify anything like `st2_version: 2.2.*`. Also, it's probably a bad show http://blog.jasonantman.com/2014/07/how-yum-and-rpm-compare-versions/ without wrapping the logic as a custom Ansible plugin with some Python.

roles/st2/tasks/main.yml
```diff
@@ -18,12 +18,25 @@
...
+  when: st2_version == "present"
+  notify:
+    - restart st2
+    - reload st2
+  tags: st2
+
 - name: Install pinned st2 package
   become: yes
   package:
     name: st2={{ st2_version }}-{{ st2_revision }}
```
armab: After check, needs an adjustement, since it's `apt` format: `apt install st2=2.2.1-5`
yum is different: `yum install st2-2.2.1-5`.

It's interesting that revision is not a requirement for yum. It's possible to do: `yum install st2-2.2.1`, while it's not possible to specify version without revision in apt.

armab:
@bigmstone FYI I reverted back to your initial commit with `present`.
Sorry for cofusion, but I initially wanted us to improve the versioning story here, which is already confusing and requires a good overhaul considering corner cases with `apt`/`yum` and package versioning.

Will see/experiment what I can do in this and next PRs.

PR #154, opened 2017-07-25
WIP: Update default node version to 6

lakshmi-kannan:
** DO NOT MERGE UNTIL FOLLOWING PR GETS MERGED ** 
https://github.com/StackStorm/st2chatops/pull/78

## Related
https://github.com/StackStorm/st2docs/pull/559
https://github.com/StackStorm/st2chatops/pull/78
https://github.com/StackStorm/st2-packages/pull/473

### DIFF
--- a/roles/nodejs/defaults/main.yml
+++ b/roles/nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "4"
+nodejs_major_version: "6"


roles/nodejs/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "4"
+nodejs_major_version: "6"
```
lakshmi-kannan: @armab When we do this, we probably need to a ansible-st2 release. I am not sure if this is the best way to do it or let people override this version in their playbooks. I am open to suggestions.

armab: We don't handle migrations or other special cases in our Ansible playbooks. This would be utterly complicated to do.

So creating new `ansible-st2` release for this change sounds good. Apart of that, let's merge this after we fully release + announce StackStorm `v2.4.0`.

PR #157, opened 2017-08-01
Install MongoDB 3.4 by default

Kami:
Part of https://github.com/StackStorm/st2/issues/3592 change.

Since we don't have per version branching in please yet, we should only merge this after v2.4.0 is out.


Kami:
test

armab:
Thanks!

Will merge it with similar `node 6` update by @lakshmi-kannan: https://github.com/StackStorm/ansible-st2/pull/154 
After that release new `ansible-st2` version.

PR #159, opened 2017-08-23
ansible-st2 readme update

AndyMoore:
make documentation more explicit to show that when the st2 version is pinned, mistral must be too


armab:
Closing as st2mistral is deprecated.

PR #160, opened 2017-08-24
Pin MongoDB and NodeJS version during install

armab:
Rely on major MongoDB and NodeJS versions during the package install.

This is a follow-up for https://github.com/StackStorm/ansible-st2/pull/154 and https://github.com/StackStorm/ansible-st2/pull/157


roles/mongodb/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+# MongoDB version to install
+# Should be '3.2' or '3.4'
+mongodb_version: "3.4"
```
armab: @cognifloyd this looks like exactly what you asked recently.

cognifloyd: Yeah. This looks very nice. It's pretty much what I did, though I still used the key directly from mongo.
Hopefully I'll have time soon to push PRs for everything else I fixed (though I don't need to push my mongodb role changes now.) :+1:

cognifloyd:
One more place where the mongo version is hardcoded:
https://github.com/StackStorm/ansible-st2/blob/efa16f3e919dffbdf51fcace46a13e9075f95f34/roles/mongodb/meta/main.yml#L3

cognifloyd:
This needs to be rebased on master to take advantage of the auto testing you just added.

armab:
@cognifloyd good catch, thanks!

roles/mongodb/tasks/mongodb_yum.yaml
```diff
@@ -1,30 +1,41 @@
...
+    baseurl: https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/{{ mongodb_major_minor_version }}/x86_64/
+    gpgkey: https://www.mongodb.org/static/pgp/server-{{ mongodb_major_minor_version }}.asc
     state: present
   tags: [databases, mongodb]
 
 - name: yum | Install mongodb
   become: yes
   yum:
-    name: mongodb-org
+    name: mongodb-org-{{ mongodb_version }}*
```
humblearner: Shouldn't this be same as: https://github.com/StackStorm/ansible-st2/pull/160/files#diff-7bdf986c5bd89f75152b5fcc42b77680R24 , explicitly install dependencies.

armab: No, I checked under CentOS and it pulls correct package version dependencies.

So there is a diff in logic for installing `deb` and `rpm` mongo packages.

armab:
Thanks @cognifloyd and @humblearner for reviewing.
Merging this.

Issue #75, opened 2017-01-09
Improve Security practices


armab:
From the: 
- https://github.com/StackStorm/st2-packages/pull/405
- https://github.com/StackStorm/st2docs/pull/363

At the moment we don't even configure any username/passwords for external services like Mongo, RabbitMQ with st2 Ansible installation. Additionally, we're not sure whether those services are listening on localhost or are open to public net.

All of that needs better control and improvement.

cognifloyd:
Here's a gist of some tasks that will idempotently (or at least I think they're idempotent):

- Turn on authorization in mongo, add admin user, and add st2 user
- Add rabbitmq management plugin, remove guest user and add a rabbitmq user for stackstorm

There's also a set of group_vars for my stackstorm group in my inventory.

https://gist.github.com/cognifloyd/c03cafc62f26a3ea0cfe1a5ed5c4238d

cognifloyd:
#162 for mongo and #164 for rabbitmq

PR #162, opened 2017-09-05
Add ability to enable mongo auth and add users

cognifloyd:
Adds auth to mongo when the `mongodb_auth_enable` flag is true.
Any additional users should be passed in via `mongodb_users`.
There are several other default variables as well (like `mongodb_host`,
`mongodb_port` which are used when logging in) that can be overridden
in the play or inventory that uses this role.
That is this PR completes this TODO from #75 for mongo security:
- [x] Configure/Set username/password

This does not attempt to generate any passwords. It only adds users if
explicitly requested, but allows external tasks/roles to import the
mongodb_auth.yml tasks to add users as required after mongo is installed.
That is, this PR skips this TODO from #75:
> ~If not explicitly set, passwords should be generated randomly and placed in st2.conf~

This uses lineinfile to edit mongod.conf with regexs to catch as many
edge cases in yaml formatting (spaces, quotes) as possible.
Also, this uses a bit of python to validate that the yaml file was
modified in such a way that it is still valid yaml, and the entries
intrduced in the file are present as expected.

This uses the mongo shell to see if authorization is required before
adding the admin user. This should handle cases such as the localhost
exception in a new install or upgrades where auth is not enabled or
upgrades where auth is enabled.

This installs pip using system packages if `pip` is not present. `pip`
is needed to install pymongo which is used to manage mongo users through
the mongodb_user ansible module.

As explained in comments, we only update mongo user passwords on_create
because that is the only way to maintain idempotency.

This should be idempotent.

TODO:

- [x] Run tests with and without mongodb auth enabled.

To configure host (in a separate PR) we need to modify `net.bindIp` in `mongod.conf` (see https://docs.mongodb.com/manual/reference/configuration-options/#net-options). I believe mongo only listens on 127.0.0.1 by default, but we might add some checks to be more explicit.
That is, this PR skips these TODOs from #75:

> ~Ensure we can configure host~
> ~Services should run on 127.0.0.1 by default~

### DIFF
--- a/roles/mongodb/defaults/main.yml
+++ b/roles/mongodb/defaults/main.yml
@@ -2,3 +2,27 @@
 # MongoDB version to install
 # Should be '3.2' or '3.4'
 mongodb_version: "3.4"
+
+# do not enable mongo auth by default. Make sure to add required users in mongodb_users if auth is enabled.
+mongodb_enable_auth: no
+
+# when adding auth, the login credentials to use
+mongodb_host: 127.0.0.1
+mongodb_port: 27017
+mongodb_admin_username: admin
+# For production use - please change the admin password!
+mongodb_admin_password: "{{ mongodb_default_admin_password }}"
+# The default is separate so the st2 role can provide a default without overriding a user provided password.
+mongodb_default_admin_password: "Ch@ngeMe"
+
+# Additional users to add.
+mongodb_users: []
+#  - db: st2
+#    username: st2mongo
+#    password: "Ch@ngeMe"
+#    roles: readWrite
+
+# whether or not to force a password update for any users in mongodb_users
+# Setting this to yes will result in 'changed' on every run, even if the password is the same.
+# See the comment in tasks/mongodb_auth.yml for more details.
+mongodb_force_update_password: no
--- a/roles/mongodb/tasks/mongodb_auth.yml
+++ b/roles/mongodb/tasks/mongodb_auth.yml
@@ -1,0 +1,136 @@
+---
+- name: Check for pip (used to install pymongo)
+  command: "{{ pip | default('pip') }} --version"
+  register: pip_check
+  changed_when: no
+  failed_when: no
+
+# It shouldn't matter if it's an old version of pip for installing pymongo, so use the system package instead of get-pip.py
+- name: Install pip with system package (used to install pymongo)
+  when: pip_check.rc != 0
+  become: yes
+  package:
+    name: 'python{% if ansible_os_family == "RedHat" and ansible_distribution_major_version == "7" %}2{% endif %}-pip'
+    state: present
+
+- name: Ensure pymongo is installed for the mongodb_user ansible module
+  become: yes
+  pip:
+    name: pymongo
+
+- name: See if mongodb authorization is enabled and users are configured
+  command: 'mongo --eval "db.getUsers()" {{ mongodb_admin_db }}'
+  # this will fail (rc 252) if auth is enabled and users are configured
+  # With the localhost exception, this will succeed (rc 0) when users are not configured even if auth is enabled.
+  # see:
+  #   - https://docs.mongodb.com/manual/core/security-users/#localhost-exception
+  #   - https://stackoverflow.com/q/31949586/1134951
+  #   - https://github.com/ansible/ansible/issues/33832#issuecomment-358031733
+  register: _mongo_authorization
+  changed_when: no # this does not change anything, it only checks
+  failed_when: no # The rc determines whether or not auth is required to create/update the admin user
+
+- name: Show mongodb auth check output
+  debug:
+    var: _mongo_authorization
+    verbosity: 2
+
+- name: Warn about default credentials
+  when: "mongodb_admin_password == 'Ch@ngeMe'"
+  debug:
+    msg: "[WARNING] Using default admin credentials for mongodb admin account! Please change them!"
+
+- name: Add mongo admin
+  mongodb_user:
+    state: present
+
+    # NOTE: on_create is idempotent - see comment below
+    update_password: on_create
+
+    name: "{{ mongodb_admin_username }}"
+    password: "{{ mongodb_admin_password }}"
+    database: "{{ mongodb_admin_db }}"
+    roles: userAdminAnyDatabase
+
+    login_host: "{{ mongodb_host }}"
+    login_port: "{{ mongodb_port }}"
+    login_user: "{{ (_mongo_authorization.rc == 0) | ternary(omit, mongodb_admin_username) }}"
+    login_password: "{{ (_mongo_authorization.rc == 0) | ternary(omit, mongodb_admin_password) }}"
+    login_database: "{{ (_mongo_authorization.rc == 0) | ternary(omit, mongodb_admin_db) }}"
+
+- name: Enable security section in mongod.conf
+  become: yes
+  lineinfile:
+    path: /etc/mongod.conf
+    regexp: |-
+      ^[#'"\s]*security['"]?\s*:
+    line: 'security:'
+    validate: |
+      {{ mongodb_python }} -c '
+      import yaml, io
+      if "security" not in yaml.safe_load(io.open("%s")):
+          exit(1)
+      '
+
+- name: Enable authentication in mongod.conf
+  become: yes
+  lineinfile:
+    path: /etc/mongod.conf
+    insertafter: '^security:'
+    # two space indentation (the default) assumed
+    line: '  authorization: enabled'
+    regexp: |-
+      ^[#'"\s]+authorization['"]?\s*:
+    validate: |
+      {{ mongodb_python }} -c '
+      import yaml, io
+      if yaml.safe_load(io.open("%s"))["security"]["authorization"] != "enabled":
+          exit(1)
+      '
+  register: _enable_mongo_auth
+  #notify: restart mongodb
+
+- name: Restart mongodb to enable auth before adding additional users
+  # This allows us to safely assume auth is already enabled when adding more users
+  when:
+    - mongodb_users
+    - _enable_mongo_auth is changed
+  become: yes
+  service:
+    name: mongod
+    state: restarted
+# TODO: Use filtered flush_handlers once this is merged: https://github.com/ansible/ansible/pull/25573
+#       As is, mongo may restart twice, once here and whenever the upgrade handlers get flushed.
+#  when: mongodb_users
+#  meta:
+#    name: flush_handlers
+#    filter: restart mongodb
+
+- name: Add additional mongo users
+  mongodb_user:
+    state: present
+
+    # NOTE: on_create is idempotent, always is not.
+    # With `update_password: on_create`, mongodb_user checks to see if the user
+    # (a) exists on the db, and (b) has the same roles,
+    # and then it only adds the user if it's not there or the roles have changed.
+    # With `update_password: always`, mongodb_user cannot tell if the password
+    # needs to be changed without attempting a login with those credentials.
+    # But mongodb_user does not currently implement such a check.
+    # A comment in mongodb_user points to https://jira.mongodb.org/browse/SERVER-22848
+    update_password: "{{ mongodb_force_update_password|ternary('always', 'on_create') }}"
+
+    name: "{{ item.username }}"
+    password: "{{ item.password }}"
+    database: "{{ item.db }}"
+    roles: "{{ item.roles|default('readWrite') }}"
+
+    login_host: "{{ mongodb_host }}"
+    login_port: "{{ mongodb_port }}"
+    login_user: "{{ mongodb_admin_username }}"
+    login_password: "{{ mongodb_admin_password }}"
+    login_database: "{{ mongodb_admin_db }}"
+  loop_control:
+    label: "{{ item.username }} on {{ item.db }}"
+  with_items: "{{ mongodb_users }}"
+
--- a/.kitchen.yml
+++ b/.kitchen.yml
@@ -55,3 +55,18 @@ platforms:
 
 suites:
   - name: default
+  - name: secure_mongo
+    provisioner:
+      extra_vars:
+        st2_config:
+          database:
+            username: st2mongo
+            password: Sw33tPassw0rd
+            db_name: st2
+        mongodb_enable_auth: yes
+        mongodb_default_admin_password: AdminP@ssw0rd
+        mongodb_users:
+          - db: st2
+            username: st2mongo
+            password: Sw33tPassw0rd
+            roles: readWrite,dbAdmin
--- a/roles/mongodb/README.md
+++ b/roles/mongodb/README.md
@@ -1,0 +1,63 @@
+Ansible Role: stackstorm.mongodb
+================================
+
+Installs MongoDB. If the `mongodb_auth_enable` boolean is enabled, then this also enables authentication, adds an admin
+user, and adds any other users defined in `mongodb_users`.
+
+Requirements
+------------
+
+Enabling mongo auth requires the pymongo python module (requirement of the mongo_users ansible module).
+
+Role Variables
+--------------
+
+These default variables can be set in the inventory's group or host vars, or pass them in as vars in the playbook that
+uses this role. An example of passing in some of these vars is shown in an example playbook below.
+
+* `mongodb_version`: The major.minor version to install (only 3.4 or 3.2 are supported).
+* `mongodb_enable_auth`: Whether or not to enable auth in mongodb (default: no)
+* `mongodb_host`: Login to this host to add users (default '127.0.0.1')
+* `mongodb_port`: Login on this port to add users (default '27017')
+* `mongodb_admin_username`: The admin's username (default 'admin')
+* `mongodb_admin_password`: The admin's password (default: generate a random password and store it in a file)
+* `mongodb_users`: A list of users to add (see example playbook below; default: [])
+* `mongodb_creds_dir`: The directory that should hold any generated credentials like admin (default: '.')
+
+Dependencies
+------------
+
+On RedHat family distributions, this depends on the stackstorm.epel role.
+
+Example Playbook
+----------------
+
+This playbook installs mongo without enabling auth or adding any users:
+
+    - hosts: localhost
+      roles:
+        - role: StackStorm.stackstorm/roles/mongodb
+
+
+This playbook installs mongo, enables auth, and adds a stackstorm user, and force update the password if it already exists:
+
+    - hosts: localhost
+      roles:
+        - role: StackStorm.stackstorm/roles/mongodb
+          vars:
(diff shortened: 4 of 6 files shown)


cognifloyd:
changed the base so that it'll pick up the travis tests... Only the last two commits are part of this PR.

cognifloyd:
squashed and rebased

cognifloyd:
TODO: Add:

cognifloyd:
Maybe add something like the following before this task https://github.com/StackStorm/ansible-st2/blob/064a4ee29c1cb986bb5a6103a653bc5be2ea0e2c/roles/st2/tasks/main.yml#L55
```yaml
  - name: Add the mongo user
    include_role:
      #name: StackStorm.mongodb
      name: mongodb
      tasks_from: mongodb_auth
    vars:
      - db: "{{ st2_mongo_db }}"
        username: "{{ st2_mongo_username }}"
        password: "{{ st2_mongo_password }}"
        roles: readWrite,dbAdmin
```

With these extra defaults in st2 role:
```yaml
st2_creds_dir: '.'
st2_mongo_db: st2
st2_mongo_username: st2mongo
st2_mongo_password: lookup('password', '{{ st2_creds_dir }}/mongodb-' + inventory_hostname + '-' + st2_mongo_username + ' length=42' )
```

And then, add something to manage setting mongo host/port (mentioned in #75)

cognifloyd:
I added a todo list based on #75

cognifloyd:
To add the mongo access settings to st2.conf:

Add the following after this task: https://github.com/StackStorm/ansible-st2/blob/064a4ee29c1cb986bb5a6103a653bc5be2ea0e2c/roles/st2/tasks/main.yml#L55
```yaml
  - name: Configure mongo auth settings in st2.conf (overrides setting in st2_config)
      become: yes
  ini_file:
    path: /etc/st2/st2.conf
    section: keyvalue
    option: encryption_key_path
    value: "{{ st2_datastore_key_file }}"
```
would this need a `notify: restart all st2` since everything uses mongo?

cognifloyd:
I could just grab the settings out of `st2_config`, but then if the user doesn't set the mongo credentials up in `st2_config` I'd need a way to merge in some default variables and then set them in st2.conf anyway. Is it better to make end users set this in the `st2_config` dict, or separate vars (since I have to have separate vars to secure it by default anyway)?

cognifloyd:
The task using the mongodb role would need some flag to disable using that role if mongo users are managed by some external role, like say in HA.

I'm afraid that adding something like this will make the roles too interdependent and prevent reusing the st2 role (eventually) in HA setups. That's why I initially left the user creation to the playbook. But, maybe that is part of what makes this feature incompatible with current st2 internal infrastructure. So, I'm looking for how to automatically add the st2 mongo user without extra intervention from the playbook author. Hopefully, that will make it easier to merge this.

roles/mongodb/defaults/main.yml
```diff
@@ -2,3 +2,25 @@
...
+mongodb_enable_auth: no
+
+# the directory used to store generated credentials in
+mongodb_creds_dir: '.'
+
+# when adding auth, the login credentials to use
+mongodb_host: 127.0.0.1
+mongodb_port: 27017
+mongodb_admin_username: admin
+mongodb_admin_password: "{{ lookup('password', '{{ mongodb_creds_dir }}/mongodb-' + inventory_hostname + '-' + mongodb_username + ' length=42' ) }}"
```
armab: It seems like not ideal option to save mongo admin password in plain text.
Need to think about the alternative approaches/directions.

roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -0,0 +1,68 @@
...
+- name: Enable authentication in mongod.conf
+  become: yes
+  lineinfile:
+    path: /etc/mongod.conf
+    insertafter: '^security:'
+    line: '  authorization: enabled'
+  notify: restart mongod
+
+# Restart mongod if required here, so that we can safely assume that auth is already enabled when adding more users.
+- meta: flush_handlers
```
armab: I would highly recommend to avoid `flush_handlers`, since it will trigger tasks in all another roles we're running.

roles/mongodb/defaults/main.yml
```diff
@@ -2,3 +2,25 @@
...
+
+# Additional users to add.
+mongodb_users: []
+#  - db: st2
+#    username: st2mongo
+#    password: "{{ lookup('password', '{{ mongodb_creds_dir }}/mongodb-' + inventory_hostname + '-' + st2mongo_username + ' length=42' ) }}"
+#    roles: readWrite
+
+# whether or not to force a password update for any users in mongodb_users
+mongodb_force_update_password: no
```
armab: Any reasons why do we need to set `update_password` to `no`?

I guess we'll see unexpected behavior while changing the password in the following example:

```yaml
          vars:
            mongodb_users:
              - db: st2
                username: st2mongo
                password: "new-password"
```

I think it's always desired to have password updated when specifying new one for the mongo user.

roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -0,0 +1,68 @@
...
+    regexp: '^#security:'
+    replace: 'security:'
+
+- name: Enable authentication in mongod.conf
+  become: yes
+  lineinfile:
+    path: /etc/mongod.conf
+    insertafter: '^security:'
+    line: '  authorization: enabled'
+  notify: restart mongod
```
armab: The correct handler name is: `restart mongodb`, according to https://github.com/StackStorm/ansible-st2/blob/master/roles/mongodb/handlers/main.yml

roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -0,0 +1,68 @@
...
+  become: yes
+  pip:
+    name: pymongo
+
+- name: See if authorization is enabled in mongod.conf
+  check_mode: yes
+  lineinfile:
+    path: /etc/mongod.conf
+    insertafter: 'security:'
+    line: '  authorization: enabled'
```
armab: I think we'll need an additional `regexp` to make the `lineinfile` less error-prone.

armab:
Also let's not rush with the random passwords for st2mongo user. Not sure yet where we should generate it: is it user's responsibility and we show how in play itself or it's a role responsibility (and it's not clear which role exactly).

Let's clarify this PR and hopefully we can add more password generation functionality as another PR.

roles/mongodb/defaults/main.yml
cognifloyd: We need to make sure `update_password: always` is idempotent. I think I was having issues with that so I switched to `update_password: no`. Re-reading the docs it looks like `always` is supposed to be idempotent, so maybe I'm remembering wrong.

I'm okay switching this to always, or to an update_password var that defaults to always.

roles/mongodb/tasks/mongodb_auth.yml
cognifloyd: Isn't it sad that there's no yaml module? Ansible speaks yaml, but it can't easily change configuration vars in yaml :(
BTW, this was fun to figure out how to check if authorization is enabled via a check_mode task.

An additional regexp? To deal with alternative indentation? By default the file uses 2-space indentation. How often do people switch away from that default?

roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -0,0 +1,68 @@
...
+    path: /etc/mongod.conf
+    regexp: '^#security:'
+    replace: 'security:'
+
+- name: Enable authentication in mongod.conf
+  become: yes
+  lineinfile:
+    path: /etc/mongod.conf
+    insertafter: '^security:'
+    line: '  authorization: enabled'
```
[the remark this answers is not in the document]
cognifloyd: see comment above.

roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -0,0 +1,68 @@
...
+    login_host: "{{ mongodb_host }}"
+    login_port: "{{ mongodb_port }}"
+    login_user: "{{ _mongo_authorization|changed | ternary(omit, mongodb_admin_username) }}"
+    login_password: "{{ _mongo_authorization|changed | ternary(omit, mongodb_admin_password) }}"
+
+- name: Enable security section in mongod.conf
+  become: yes
+  replace:
+    path: /etc/mongod.conf
+    regexp: '^#security:'
```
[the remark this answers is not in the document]
cognifloyd: If the `regexp` doesn't match and the `replace` line doesn't exist in the file, then it adds that `replace` line at the end of the file. So, I'm not relying on `#security:`, I'm making sure to remove the commented one if it exists, and I'm inserting the security section in place of the comment when it exists. If it doesn't, then the section ends up at the end of the file. Then, in the next task, I add authorization right after the security section head.

roles/mongodb/defaults/main.yml
cognifloyd: https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/database/mongodb/mongodb_user.py#L440-L455
mongodb_user is not idempotent. If you do update_password: always then every time that task runs it will report "changed" even if the password is the same as what is already there.
If you do update_password: on create, then the module checks to see if the user (a) exists on the db, and (b) has the same roles, and then it only adds the user if it's not there or the roles have changed.
So, we can do update_password: always only if we're ok with having continuously report changed on every run.

cognifloyd: There is a comment in the module about checking the password:
```python
            # Here we can  check password change if mongo provide a query for that : https://jira.mongodb.org/browse/SERVER-22848
            #newuinfo = user_find(client, user, db_name)
            #if uinfo['role'] == newuinfo['role'] and CheckPasswordHere:
            #    module.exit_json(changed=False, user=user)
```
In that [mongo Jira issue](https://jira.mongodb.org/browse/SERVER-22848), they prefer:
> all password checks should always be audited as login attempts for security purposes

So, we to try and make this more idempotent we can try and login with the username/password to see if a password update is required. That will have the side effect of having a login attempt logged every time we run the playbook. If that's ok, great. We can either do that login attempt by

1. Change the module to make it do the test login.
2. Add some set of tasks to the playbook that try to login before adding the user... I don't know what that would look like. Maybe we can use the [mongodb lookup](https://docs.ansible.com/ansible/devel/plugins/lookup/mongodb.html) to login and read something (what? How does login failure via lookup act?) Then, if that login was successful, use `update_password: on create` or if unsuccessful (user doesn't exist or password failed) then `update_password: always`.

roles/mongodb/tasks/mongodb_auth.yml
cognifloyd: There's a yaml editing module, but I hesitate to rely on it:
https://github.com/kwoodson/yedit

- build system feels weird: They are somehow squashing several files to generate the module, but they distribute all of the pieces with the finished module
- packaging feels weird: They're distributing the module inside of a role which is very non-standard
- distribution is difficult: It's not available on ansible-galaxy, so is it really ready for third party use? Do I really want to maintain the distribution of someone else's module for use in my playbook when they won't even bother to publish it on galaxy?

cognifloyd: Other approaches:
 - [this one](https://galaxy.ansible.com/greendayonfire/mongodb/) uses a template to create a temporary (templated) mongod.conf with authentication disabled while adding the admin and other users, start mongo with that config, then rebuild the real config with another template.
- [this one](https://galaxy.ansible.com/Stouts/mongodb/) uses a template for the config file that is an ini looking thing instead of yaml.
- others I looked at use templates for the conf file.

I don't want to use a template unless we're willing to control the whole conf file.

cognifloyd:
Maybe we need an st2data or st2init role to handle initialization of the
various ST2 required services (mongo, rabbitmq, postgres). Each services
user and data initialization tasks would be in a separate file (mongo.yml,
rabbitmq.yml, postgres.yml) and included in main.yml to provide default ST2
behavior. This would allow customizable behavior because a playbook could
import only one of those files.

roles/mongodb/defaults/main.yml
armab: Thanks for the research, that makes sense :+1: 
Would be good to include a code comment for future contributors about the idempotence failure in Ansible module, so we won't try to change this again.

roles/mongodb/tasks/mongodb_auth.yml
armab: I still think adding a `regexp` here will make it less error prone. Like not perfect, but much better and good enough without going into overcomplication.

Consider the following examples: 
```
  authorization: enabled 
   authorization: enabled
  authorization: enabled
  authorization:enabled
  authorization: "enabled"
  authorization: 'enabled'
  authorization : enabled
```

cognifloyd: Ah. I see. Makes sense. I add a regex then.

cognifloyd: https://github.com/StackStorm/ansible-st2/pull/162

cognifloyd: Ah. I see. I will add the regex then :)

cognifloyd:
rebased

roles/mongodb/tasks/mongodb_auth.yml
cognifloyd: Here's hope for a better way to do this in the future: https://github.com/ansible/ansible/pull/25573

cognifloyd:
Re: Flushing handlers:
Here's hope for a better way to do this in the future: ansible/ansible#25573

Re: Where to generate passwords / Who is responsible / Saving in plain text:
I don't know yet.

roles/mongodb/tasks/mongodb_auth.yml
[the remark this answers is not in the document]
cognifloyd: I replaced this line with a lineinfile and a validate python script that makes sure the yaml is valid after modification and has the values we added to it.

roles/mongodb/tasks/mongodb_auth.yml
cognifloyd: Replaced this with an explicitly restart mongodb task if the previous task changed something. I left a TODO to revisit this when flush_handlers can take a filter.

cognifloyd:
> Also let's not rush with the random passwords for st2mongo user.

I removed random password generation, using "Ch@ngeMe" with a warning if the user doesn't change it.

roles/mongodb/defaults/main.yml
cognifloyd: dropped the password generation for now.

cognifloyd:
I added an st2 task that adds the mongo user (and enables st2 auth) based on the contents of `st2_config`. If `st2_config.database.username` and `st2_config.database.password` are present (not sure what should happen if only username or password is defined), then enable the mongo auth and configure the mongo user. Plus, I use the `st2_config.database.password` as a default admin password if the user has not set `mongodb_admin_password`.

cognifloyd:
@armab So, I got something of a test to try the mongo tasks in travis, but `pip`.

I used `pip` to install `pymongo` because the system packages in CentOS are really old.
However, `pip` is not installed on travis by default. So, how would you like to approach that?
I've used `ansiblebit.pip` before (it looked like one of the best maintained pip roles). It supports all of StackStorm's platforms too. What would you think of adding [`ansiblebit.pip`](https://github.com/ansiblebit/pip)? 
```yaml
- name: Install pip
  include_role:  # include so that the mongodb_auth.yml file can be included not just imported.
    name: ansiblebit.pip
```

Second issue:
I'm not familiar with kitchen... I'm trying to hack in a second suite, but I don't know how merging of global `provisioner` settings works with `suite.provisioner` settings. For instance, do I need to specify the entire contents of `extra_vars` or just the additional `st2_config` var?

cognifloyd:
squashed into logical commits. Now just waiting for the tests to pass both with and without mongo auth enabled.

cognifloyd:
DONE! No longer WIP. Travis is now testing two suites: one without auth enabled, and one with auth enabled. I think this is ready for another review, and hopefully, a merge @armab.

cognifloyd:
Rebased and squashed various commits to make it easier to review each piece/commit of this PR.

cognifloyd:
K. I have a modified version of this that drops the `requirements` and uses system packages to install `pip` if needed. I will rebase on top of master and push those changes once #200 is merged. That brings the commit count down to 3.

cognifloyd:
Dropped changes to `st2` role. Now adding users is ONLY activated if the `mongodb_enable_auth` is true (defaults to false).

cognifloyd:
When you get to reviewing, I recommend reviewing files in this order (simplest first):

cognifloyd:
When you review, I recommend this order (which puts the simplest changed files first):

- Part of the [first commit]() (actual functionality and docs):
  - [roles/mongodb/tasks/main.yml](https://github.com/StackStorm/ansible-st2/pull/162/commits/a395b36720b64c7a0dc3fa8b37b17ef5ce4d3033#diff-6ced052f1ce1651196fdcca20576ce29) _include when: `mongodb_enable_auth: yes`_
  - [roles/mongodb/vars/main.yml](https://github.com/StackStorm/ansible-st2/pull/162/commits/a395b36720b64c7a0dc3fa8b37b17ef5ce4d3033#diff-906964f349f13c21eadc8afa968ba387) _vars: `mongodb_python` used in `validate`_
  - [roles/mongodb/defaults/main.yml](https://github.com/StackStorm/ansible-st2/pull/162/commits/a395b36720b64c7a0dc3fa8b37b17ef5ce4d3033#diff-afb6ffa2a479b6507d3a2171a7fa08e5) _various generic variables that users might wish to change_
  - [roles/mongodb/tasks/mongodb_auth.yml](https://github.com/StackStorm/ansible-st2/pull/162/commits/a395b36720b64c7a0dc3fa8b37b17ef5ce4d3033#diff-582215a1d16dc615f8c0321e9e1059ca) _logic to enable mongodb auth._
    - This must include:
        1. enable `security.authentication`;
        2. add an admin user to admin db;
        3. add regular users.
  - [roles/mongodb/README.md](https://github.com/StackStorm/ansible-st2/pull/162/commits/a395b36720b64c7a0dc3fa8b37b17ef5ce4d3033#diff-99514086d5b7b8cc7dbc55b611d41868) _document vars specific to the mongodb role_
- The only file in the [second commit] (add secure test suite):
  - [.kitchen.yml](https://github.com/StackStorm/ansible-st2/pull/162/commits/3b9ed185c62a5d491d82ee928981cf0a60b64246#diff-d3fbd21bc6be7ce4531c27a82b2619de) _`suite.provisioner.extra_vars` gets merged with global level `provisioner.extra_vars`_

cognifloyd:
I will try to break this down into ity-bity PRs. Starting with a way to install pip #204.
I'll close this for now since there's no resources to review it.

PR #163, opened 2017-09-05
FEATURE: Allow for installing rabbitmq_plugins

### LINKED ISSUE #147: Configure RabbitMQ plugins
Ability to install/enable custom RabbitMQ plugins.

For example, REST API plugin: https://www.rabbitmq.com/management.html

cognifloyd:
Resolves #147 

rabbitmq_plugins is a comma separated list of plugins to add (existing plugins are not removed).
By default no additional plugins are installed.


roles/rabbitmq/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# set to [] to disable.
+rmq_plugins:
```
armab: Please prefix the var by role name to follow the best practices, eg `rabbitmq_plugins`

roles/rabbitmq/tasks/main.yml
```diff
@@ -15,3 +15,13 @@
     enabled: yes
     state: started
   tags: rabbitmq
+
+- name: Add RabbitMQ plugins
+  become: yes
+  rabbitmq_plugin:
+    name: "{{ item }}"
+    state: enabled
+    prefix: "{{ is_redhat6 | ternary(rmq_redhat6_prefix, omit) }}"
```
armab: Any comment/explanation would be nice to understand why `prefix` is needed on el6-based systems

roles/rabbitmq/tasks/main.yml
```diff
@@ -15,3 +15,13 @@
...
+  with_items: "{{ rmq_plugins }}"
+  when: rmq_plugins
+  tags: rabbitmq
```
armab: Do we need to restart rabbitmq once the plugin is added?

roles/rabbitmq/tasks/main.yml
```diff
@@ -15,3 +15,13 @@
     enabled: yes
     state: started
   tags: rabbitmq
+
+- name: Add RabbitMQ plugins
+  become: yes
+  rabbitmq_plugin:
+    name: "{{ item }}"
```
armab: From http://docs.ansible.com/ansible/latest/rabbitmq_plugin_module.html I see the `names` (not `name`) is comma-separated and there is additional `new_only` option which as I understand might disable previously enabled plugins.

So this logic needs to be double-checked.

roles/rabbitmq/vars/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+rmq_redhat6_prefix: "/usr/lib/rabbitmq/"
+is_redhat6: "{{ (ansible_os_family == 'RedHat' and ansible_distribution_major_version == '6') }}"
```
armab: Same as before, let's prefix all vars with the role name.

roles/rabbitmq/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
...
+  # you probably want at least one user tagged with administrator to use this plugin
+  - rabbitmq_management
```
armab: Still not decided/sure if we should ship the role with this plugin enabled or not (empty `[]`).
Theoretically, installing additional plugin `rabbitmq_management` by default might be not desired/expected by users behavior.

roles/rabbitmq/tasks/main.yml
cognifloyd: I don't think so:

> The rabbitmq-plugins comand will enable or disable plugins by updating the plugin configuration file. It will then contact the running server to tell it to start or stop plugins as needed. You can use the  -n option to specify a different node, or use  --offline to only change the file.
https://www.rabbitmq.com/plugins.html

I read that to mean that a running rabbitmq both loads/unloads the plugin without a restart (as long as the --offline flag isn't used. The rabbitmq_plugin module doesn't have a way to include the offline flag, so we shouldn't ever need a restart when adding/removing plugins.

roles/rabbitmq/tasks/main.yml
cognifloyd: The rabbitmq RPM in EL 6 systems only includes the rabbitmq-plugins binary at `/usr/lib/rabbitmq/bin/rabbitmq-plugins` (see the list of files in https://centos.pkgs.org/6/epel-x86_64/rabbitmq-server-3.1.5-1.el6.noarch.rpm.html). In the EL 7 RPM, the rabbitmq-plugins binary is also available via `/usr/sbin/rabbitmq-plugins` (see the list of files in https://centos.pkgs.org/7/epel-x86_64/rabbitmq-server-3.3.5-34.el7.noarch.rpm.html). As the binary is already on the path in EL7, but not in EL6, we have to use a prefix to add `/usr/lib/rabbitmq/bin/` to the search path in EL6.

I will try write something more concise to leave a comment in the playbook.

roles/rabbitmq/tasks/main.yml
cognifloyd: ok. `name` is an alias for `names`, but I guess it would make sense to use the csv list instead of using an ansible with loop. I didn't see `new_only`: I think that should definitely be on by default.

cognifloyd:
I agree. Shipping with the management plugin enabled by default might not be desirable, because you also have to coordinate adding the administrator tag to some user to make it useful. Might as well leave it disabled and let people turn it on as needed (I turn it on).

I also changed the var names and added the new_only flag to ensure other plugins aren't removed. Nothing else needs to be restarted either.

roles/rabbitmq/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+# A comma separated list of plugins to enable (existing plugins will not be disabled)
+rabbitmq_plugins: ''
```
armab: I'm +1 for a list as it was before (to follow the yaml data types), we just need to transform it afterwards in `rabbitmq_plugin`.

Also what will happen if on first deployment we have `rabbitmq_plugins: [rabbitmq_management]` and on second deployment we remove it: `rabbitmq_plugins: []`.
Will that automatically disable unneeded plugins for us?

cognifloyd: OK. I will turn it into a yaml list, and then join(',') it in the play.

roles/rabbitmq/tasks/main.yml
```diff
@@ -15,3 +15,14 @@
     enabled: yes
     state: started
   tags: rabbitmq
+
+- name: Add RabbitMQ plugins
+  become: yes
+  rabbitmq_plugin:
+    names: "{{ rabbitmq_plugins }}"
+    state: enabled
+    new_only: yes
```
cognifloyd: > Also what will happen if on first deployment we have rabbitmq_plugins: [rabbitmq_management] and on second deployment we remove it: rabbitmq_plugins: [].
Will that automatically disable unneeded plugins for us?

With "new_only", it will only ever add new plugins. It won't remove old ones. I could add a `rabbitmq_remove_unlisted_plugins` var (unsure on that verbose name) that would do `new_only: no` when specifically requested, but default to `yes`. Maybe it would be clearer to use `rabbitmq_plugins_new_only` to merely pass through the bool, but default it to yes.

cognifloyd: done. I used the simpler `rabbitmq_plugins_new_only` var.

armab:
Looks good, will test it tomorrow manually.

roles/rabbitmq/tasks/main.yml
```diff
@@ -15,3 +15,14 @@
     enabled: yes
     state: started
   tags: rabbitmq
+
+- name: Add RabbitMQ plugins
+  become: yes
+  rabbitmq_plugin:
+    names: "{{ rabbitmq_plugins|join(',') }}"
+    state: enabled
+    new_only: "{{ rabbitmq_plugins_new_only }}"
```
armab: After thinking a while, personally I'd prefer to hardcode this to `no` and remove var `rabbitmq_plugins_new_only` for simplicity reasons.

This way Ansible will manage full state of rabbitmq plugins.

WDYT? For your infra, do you need this option as `yes`? I'm OK to leave that as is then.

IMO if we're in infra-as-a-code field, the point of truth is configuration management tool and if any custom plugin is installed manually or by any other system - it should be removed by Ansible to keep in sync with the git/code.

cognifloyd: For my infra where everything is running on one machine, I don't care. But, will people want this for a HA setup? I can imagine having multiple playbooks logging in to a rabbitmq cluster to set up the users, vhosts, and plugins that are expected for a single app. But you wouldn't want one app removing the info for another app. So, I would err on the side of only removing if specifically requested for multi-tenant rabbitmq HA installs.

armab: At the moment we don't focus on any HA-deployment cases for `rabbitmq` role.
See https://github.com/StackStorm/ansible-st2/issues/119 similar discussion.

The current position about `ansible-st2` is that we provide minimal installer similar by logic to `curl|bash` demo installer.
Having said that, at the same time we leave an option to use external roles from galaxy to configure mongo or rabbitmq or postgresql how users want it.

cognifloyd: The newest commit hard codes `new_only: no`.

PR #164, opened 2017-09-05
[WIP] Add RabbitMQ users/vhosts

cognifloyd:
If `rmq_users` is defined, then the guest user is removed, and the provided users are added.

TODO:
- [ ] Add test suite to test with rabbitmq auth enabled (probably one suite for all auth including mongo)
- [ ] Be ok with increasing minimum ansible version to 2.4 (for urlsplit filter use in st2 module to extract user, vhost from `st2_config.messaging.url`).

TODO for #75 for RabbitMQ security:
- [x] Configure/Set username/password

To configure the host (in a separate PR) we will need to modify or template `/etc/rabbitmq/rabbitmq.config` (see: https://www.rabbitmq.com/networking.html#interfaces)
- [-] ~Ensure we can configure host~
- [-] ~Services should run on 127.0.0.1 by default~

Leave pw generation to the playbook user.
- [-] ~If not explicitly set, passwords should be generated randomly and placed in st2.conf~

### DIFF
--- a/roles/rabbitmq/tasks/main.yml
+++ b/roles/rabbitmq/tasks/main.yml
@@ -31,3 +31,11 @@
     prefix: "{{ rabbitmq_on_el6 | ternary(rabbitmq_el6_prefix, omit) }}"
   when: rabbitmq_plugins
   tags: rabbitmq
+
+- name: Add/Remove RabbitMQ vhosts
+  when: rabbitmq_vhosts or rabbitmq_absent_vhosts
+  include: rabbitmq_vhosts.yml
+
+- name: Add/Remove RabbitMQ users
+  when: rabbitmq_users or rabbitmq_absent_users
+  include: rabbitmq_users.yml
--- a/roles/rabbitmq/README.md
+++ b/roles/rabbitmq/README.md
@@ -1,0 +1,72 @@
+Ansible Role: stackstorm.rabbitmq
+=================================
+
+Installs RabbitMQ. If `rmq_users` is defined, then the guest user is removed, and the provided users are added.
+
+Requirements
+------------
+
+No additional requirements.
+
+Role Variables
+--------------
+
+These default variables can be set in the inventory's group or host vars, or pass them in as vars in the playbook that
+uses this role. An example of passing in some of these vars is shown in an example playbook below.
+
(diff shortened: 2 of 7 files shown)


roles/rabbitmq/tasks/main.yml
```diff
@@ -25,3 +25,39 @@
...
   tags: rabbitmq
+
+- name: Remove the guest user from RabbitMQ
+  become: yes
+  rabbitmq_user:
+    user: guest
+    state: absent
+  when:
+    - rmq_users|length > 0
+    - "'guest' not in rmq_users|map(attribute='user')"
```
cognifloyd: map requires jinja 2.7

roles/rabbitmq/tasks/main.yml
```diff
@@ -25,3 +25,39 @@
...
+  tags: rabbitmq
+
+- name: Add RabbitMQ vhosts
+  become: yes
+  rabbitmq_vhost:
+    vhost: "{{ rmq_vhost }}"
+    state: present
+  loop_control:
+    loop_var: rmq_vhost
+  with_items: "{{ lookup('flattened', rmq_users|map(attribute='permissions')|list ) | map(attribute='vhost')|list|unique }}"
```
cognifloyd: map requires jinja2.7.

cognifloyd:
Another option would be to require an extra play:
```
- hosts: localhost
  tasks:
    - name: Upgrade ansible host's Jinja2 when using RedHat 6
      when: ansible_os_family == "RedHat" and ansible_distribution_major_version == "6"
      pip:
        name: jinja2
```

cognifloyd:
rebased

cognifloyd:
To get that map like functionality, I could use json_query, but that would add a dep on jmespath.

cognifloyd:
So far, this only adds the ability to add the user, but the playbook user has to actually add it. So, to accomplish #75 for RabbitMQ, this needs to switch to adding the user by default, and make the host configurable. That's a much bigger change, but maybe then it'll actually get merged.

cognifloyd:
rebased

cognifloyd:
Closing for now. Once there are resources to review something, I'll deal with it again.

PR #165, opened 2017-09-05
Configure datastore encryption

### LINKED ISSUE #153: Feature Request: configure datastore secret encryption key
See: https://docs.stackstorm.com/datastore.html#securing-secrets-admin-only

cognifloyd:
Resolves #153 

This automatically configures the datastore encryption key.


cognifloyd:
This is now part of the st2 role.

roles/st2/tasks/datastore.yaml
```diff
@@ -0,0 +1,40 @@
+---
+- name: Add the st2 system service user to the service group
+  become: yes
+  user:
+    user: "{{ st2_system_service_user }}"
+    groups: "{{ st2_system_service_group }}"
```
armab: Creation of `st2` user and group is already done in packaging so this block is not needed.

roles/st2/vars/main.yml
```diff
@@ -12,3 +12,5 @@ st2_services:
   - st2api
   - st2stream
   - st2auth
+
+st2_datastore_key_file: /etc/st2/keys/datastore_key.json
```
armab: Would be nice to have a short code comment for this var.

roles/st2/tasks/datastore.yaml
```diff
@@ -0,0 +1,40 @@
...
+    append: yes
+
+- name: Make stackstorm encryption keys directory
+  become: yes
+  file:
+    path: "{{ st2_datastore_key_file | dirname }}"
+    state: directory
+    mode: 0750
+    owner: root
+    group: "{{ st2_system_service_group }}"
```
armab: Just hardcoding to `st2` would be sufficient, because there is no other option for now (st2 service user is not configurable)

roles/st2/tasks/datastore.yaml
```diff
@@ -0,0 +1,40 @@
...
+  become: yes
+  user:
+    user: "{{ st2_system_service_user }}"
+    groups: "{{ st2_system_service_group }}"
+    append: yes
+
+- name: Make stackstorm encryption keys directory
+  become: yes
+  file:
+    path: "{{ st2_datastore_key_file | dirname }}"
```
armab: Nice trick 👍

roles/st2/defaults/main.yml
```diff
@@ -34,3 +34,9 @@ st2_auth_username: testu
 st2_auth_password: testp
 # Save credentials in ~/.st2/config file
 st2_save_credentials: yes
+
+# System service user installed by the st2 packages
+st2_system_service_user: st2
+# System service user group
+st2_system_service_group: st2
```
armab: Because it's hardcoded in `deb`/`rpm` packaging, we shouldn't define these 2 variables.

roles/st2/tasks/datastore.yaml
```diff
@@ -0,0 +1,40 @@
...
+    group: st2
+
+- name: Configure encryption key in st2.conf
+  become: yes
+  ini_file:
+    path: /etc/st2/st2.conf
+    section: keyvalue
+    option: encryption_key_path
+    value: "{{ st2_datastore_key_file }}"
+  notify: restart st2api
```
armab: @bigmstone Do you remember which service (or services?) is responsible for picking up datastore encryption key?

In our infra I see we just do a `reload`, without service restart.

roles/st2/tasks/datastore.yaml
cognifloyd: Then maybe there was a bug, because the st2 user was not a member of the st2 group, so I added this to add the user to the group. The user already existed. Maybe this issue has been addressed elsewhere?

roles/st2/tasks/datastore.yaml
cognifloyd: I copied the restart from the st2bootstrap script
```
  sudo st2ctl restart-component st2api
```
https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el6.sh#L492
https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-el7.sh#L501
https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L466

roles/st2/tasks/datastore.yaml
armab: This works for me on Ubuntu16:
```
groups st2
st2 : st2 st2packs
```
`st2` user was associated with `st2` group. I'm running latest s2 2.4

Do you see any diff on EL/CentOS systems? Must be a bug then.

cognifloyd: Reading the rpm spec, it should have created the group. I don't know what I saw as it was a while ago. I'll just delete this for now, and if it pops up again, I'll file a bug against the st2-packages.

cognifloyd:
I made the requested changes. This is ready for another review @armab.

roles/st2/tasks/main.yml
```diff
@@ -60,6 +60,9 @@
     loop_var: _conf_section
   tags: st2, config
 
+- name: Configure the StackStorm datastore
+  include: datastore.yaml
+
```
armab: Just a final cosmetic request: please include `tags` for this section as used in other blocks.

cognifloyd: done

cognifloyd:
I added the tags, and tests pass.

PR #167, opened 2017-09-05
Prevent output of secrets and fix conditional

cognifloyd:
Clean up the output of playbook runs.

Prevent sending secrets to stdout (at least through the default ansible output).

Also, ansible complained about the when conditional that had the jinja template indicators `{{ }}`, but they shouldn't be used for when conditionals. This fixes that as well.


roles/st2web/tasks/certificate.yml
```diff
@@ -27,6 +27,7 @@
       path: /etc/ssl/st2/st2.crt
     - cert: "{{ st2web_ssl_certificate_key }}"
       path: /etc/ssl/st2/st2.key
+  no_log: yes
```
armab: 👍

roles/st2/tasks/main.yml
```diff
@@ -48,7 +48,7 @@
 
 - name: Configure StackStorm authentication
   include: auth.yml
-  when: "{{ st2_config.auth.enable|default(st2_auth_enable) if st2_config.auth is defined else st2_auth_enable }}"
+  when: _st2_config_auth_enable|default(st2_auth_enable, true)|bool
```
armab: Not sure if `default(st2_auth_enable, true)` is desired, comparing to previous `default(st2_auth_enable)`.

Also, talking about the refactoring, here is a shorter example:
```
when: (st2_config.auth|default({})).enable|default(st2_auth_enable)
```
so we don't create additional `_st2_config_auth_enable` in vars and keep it not that long & minimal as possible.

armab: Note: would be nice to include a small code comment about this `when` block.

roles/st2mistral/tasks/config.yml
```diff
@@ -9,6 +9,8 @@
   with_dict: "{{ _conf_options }}"
   loop_control:
     loop_var: _conf_option
+    # prevent logging passwords, auth URIs, and secrets
+    label: "conf | {{ _conf_section_name }}.{{ _conf_option.key }}"
```
armab: I'm not quite sure if we should really hide _all_ the `conf` settings in Ansible debugging to mask several secret fields.

The places we've added `no_log` are where tokens could leak during the Travis CI builds.

roles/st2mistral/tasks/main.yml
```diff
@@ -69,6 +69,8 @@
     option: connection
     value: postgresql://{{ st2mistral_db_username }}:{{ st2mistral_db_password }}@localhost/{{ st2mistral_db }}
     backup: yes
+  # no_log to prevent plain text logging of the password
+  no_log: yes
```
armab: 👍

roles/st2/tasks/main.yml
cognifloyd: I tried a lot of different things to fix this. That big ugly `_st2_config_auth_enable` is the way I found to fix it. The ternary check that is currently there never could have worked because that's not a proper jinja ternary.

If `when: (st2_config.auth|default({})).enable|default(st2_auth_enable)` works, then perfect, let's use that.

In any case, I was trying to cleanup the output so I didn't have so many errors about "{{ }}" on a when block... It turns out the entire when conditional was not correct, but that's what I'm trying to cleanup.

roles/st2mistral/tasks/config.yml
cognifloyd: Yeah, I was concerned about doing that too. This was easy though.

Do you know if `no_log` applies on the loop as a whole, or on each item of the loop? If it does, then we could check and see if a particular setting is in a blacklist, and then `no_log` it.

Note that merely changing the label does not prevent logging the output if `--diff` or a higher verbosity is enabled. `no_log` is a better solution, but I didn't want to enable no_log on all of those, because I like to use `--diff`.

roles/st2/tasks/main.yml
cognifloyd:
```
ansible all -i "localhost," -m debug -a 'msg="{{ st2_config.auth|default({}).enable|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{},"st2_auth_enable":'asdf'}'

Comment (cognifloyd) [reply roles/st2/tasks/main.yml]:
```ansible all -i "localhost," -m debug -a 'msg="{{ st2_config.auth|default({}).enable|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{},"st2_auth_enable":'asdf'}'

Comment (cognifloyd) [reply roles/st2/tasks/main.yml]: This when condition does not work.
```
$ ansible all -i "localhost," -m debug -a 'msg="{{ st2_config.auth|default({}).enable|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{},"st2_auth_enable":'asdf'}'
localhost | FAILED! => {
    "failed": true, 
    "msg": "template error while templating string: expected token 'end of print statement', got '.'. String: {{ st2_config.auth|default({}).enable|default(st2_auth_enable) }}"
}

$ ansible all -i "localhost," -m debug -a 'msg="{{ st2_config.auth|default({})['enable']|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{},"st2_auth_enable":'asdf'}'
localhost | FAILED! => {
    "failed": true, 
    "msg": "template error while templating string: expected token 'end of print statement', got '['. String: {{ st2_config.auth|default({})[enable]|default(st2_auth_enable) }}"
}
```

Comment (cognifloyd) [reply roles/st2/tasks/main.yml]: This works for the empty `{}` case.
```
ansible all -i "localhost," -m debug -a $'msg="{{ st2_config.auth|default({})|attr(\'enable\')|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{},"st2_auth_enable":'asdf'}' 
localhost | SUCCESS => {
    "msg": "asdf"
}
```
but not if it's populated:
```
ansible all -i "localhost," -m debug -a $'msg="{{ st2_config.auth|default({})|attr(\'enable\')|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{"auth":{"enable":"qwer"}},"st2_auth_enable":'asdf'}'
localhost | SUCCESS => {
    "msg": "asdf"
}
```

That's because we need to get an item from the dictionary not an attribute. I don't see a filter for getting something out of a dictionary returned by another filter.

Comment (cognifloyd) [reply roles/st2/tasks/main.yml]: Aargh. I missed you're extra parentheses. This works (just like you had it):
```
$ ansible all -i "localhost," -m debug -a 'msg="{{ (st2_config.auth|default({})).enable|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{},"st2_auth_enable":'asdf'}'
localhost | SUCCESS => {
    "msg": "asdf"
}

$ ansible all -i "localhost," -m debug -a 'msg="{{ (st2_config.auth|default({})).enable|default(st2_auth_enable) }}"' --extra-vars='{"st2_config":{"auth":{"enable":"qwer"}},"st2_auth_enable":'asdf'}'
localhost | SUCCESS => {
    "msg": "qwer"
}
```

OK. I will add this in a few minutes.

**Comment by cognifloyd** [comment]
@armab How does this look for no_log on the config options?
I just ran this on my st2 host (a CentOS 7 host) and it successfully masks only the settings that are in the `st2_config_no_log` or `st2mistral_config_no_log` vars.

**Comment by armab** [comment]
The idea of blacklisting specific config options looks good to me 👍

File: roles/st2/tasks/main.yml [hunk 5]
```diff
@@ -57,6 +58,8 @@
   with_dict: "{{ st2_config }}"
   loop_control:
     loop_var: _conf_section
+    # prevent logging passwords, auth URIs, and secrets
+    label: "conf | {{ _conf_section.key }}"
```
armab: Do we still need this part, having `st2_config_no_log`?

cognifloyd: I think so, because the value for this option includes the entire unfiltered sub dictionary of st2_config.

armab: Just to understand, where it leaks? Debug mode?
Additionally, what `conf` does in this construction: `conf | {{ _conf_section.key }}`?

cognifloyd: ~It might even be good to no_log this, as it is covered in the output of the included file~
Actually this doesn't do anything. All I get is "included: /opt/theatro/etc/ansible/roles/StackStorm.stackstorm/roles/st2/tasks/config.yml for harmony" over and over whether or not this line is here. I'll drop it.

PR #171, opened 2017-09-20
Update roles path in example ansible.cfg when installing StackStorm from galaxy

armab:
Update the path in example `ansible.cfg` to point to `StackStorm.stackstorm` when installing from Galaxy https://galaxy.ansible.com/StackStorm/stackstorm/


armab:
Thanks for making a second look at it, @cognifloyd!

PR #173, opened 2017-09-27
Prefix roles with StackStorm galaxy org

cognifloyd:
Part of #12 and #45


cognifloyd:
I thought about renaming all of the files in the roles to have a yaml extension instead of yml, as I'm renaming all of them any way. I haven't done it yet. Do you care about the mixed extensions?

armab:
Yes, file extension renaming to one common format would be good, but as part of another PR to merge it quickly.
I would like `.yml`, instead of `.yaml` as that's traditional format Ansible prefer in their docs.

cognifloyd:
Should the `.yaml`=>`.yml` change happen before or after this one?

cognifloyd:
Added #174 for yml file extension, and rebased #173 (this one) on top of it.

cognifloyd:
rebased on master

cognifloyd:
rebased

cognifloyd:
rebased

cognifloyd:
As mentioned in https://github.com/StackStorm/ansible-st2/issues/45#issuecomment-422227473, it looks like ansible will end up getting support for multi-role repos through [ansible/mazer](https://github.com/ansible/mazer). No idea when that will land, but I'm not going to keep maintaining this PR.

armab:
Same thing with Travis build failing. I'm guessing something changed on Travis side which stopped sharing secrets with the forked PRs?

Will research more.

LindsayHill:
@armab took a really brief look at this yesterday. See https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings

> Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.

Not a Travis expert though, don't know if there are other options we can use here

armab:
Merged with some TravisCI workarounds,
Thanks @cognifloyd for working on this!

PR #175, opened 2017-10-02
Remove BWC RBAC directories creation

armab:
The directory creation and permissions/ownership is done by the deb/rpm packaging now.

The build will :white_check_mark: once the change is merged in upstream.


armab:
All tests are green now, after merging respective change in `deb`/`rpm` packaging.

PR #177, opened 2017-10-03
Make sure mistral db gets upgraded too

cognifloyd:
When upgrading mistral, the database needs to be updated as well.
This adds the version number to the signal file so that on upgrade,
it will run the 'upgrade head' and 'populate' commands again.

### DIFF
--- a/roles/st2mistral/tasks/main.yml
+++ b/roles/st2mistral/tasks/main.yml
@@ -13,6 +13,7 @@
     name: st2mistral
     state: latest
   when: st2mistral_version == "latest"
+  register: mistral_install_latest
   tags: [st2mistral, skip_ansible_lint]
 
 - name: Install present st2mistral package, no auto-update
@@ -21,6 +22,7 @@
     name: st2mistral
     state: present
   when: st2mistral_version == "present"
+  register: mistral_install_present
   tags: st2mistral
 
 - name: Install pinned st2mistral package
@@ -31,7 +33,19 @@
   when:
     - st2mistral_version != "latest"
     - st2mistral_version != "present"
+  register: mistral_install_tagged
   tags: st2mistral
+
+- name: Enable mistral db upgrade
+  become: yes
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
+    - /etc/mistral/mistral-db-manage.populate.ansible.has.run
+  when: mistral_install_latest|changed or mistral_install_present|changed or mistral_install_tagged|changed
+  tags: st2mistral, skip_ansible_lint
 
 - name: Deploy database init script
   become: yes


roles/st2mistral/tasks/main.yml
```diff
@@ -85,18 +85,18 @@
 
 - name: Setup mistral DB tables, etc
   become: yes
-  shell: /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head && touch /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
+  shell: /opt/stackstorm/mistral/bin/mistral-db-manage --config-file /etc/mistral/mistral.conf upgrade head && touch /etc/mistral/mistral-db-manage.upgrade.head.{{ st2mistral_version }}.ansible.has.run
   args:
-    creates: /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
+    creates: /etc/mistral/mistral-db-manage.upgrade.head.{{ st2mistral_version }}.ansible.has.run
```
armab: With the `st2mistral_version: latest` it will still run only once.

So different mechanism is needed.

cognifloyd: True. Hmm...

armab: I think the initial approach @bigmstone suggested, - to delete the lock file once the `install st2mistral` task has been changed.

cognifloyd:
How about this, deleting the files whenever the package task changes something?

roles/st2mistral/tasks/main.yml
```diff
@@ -31,8 +33,18 @@
...
 
+- name: Enable mistral db upgrade
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
+    - /etc/mistral/mistral-db-manage.populate.ansible.has.run
+  when: mistral_install_latest|changed or mistral_install_present|changed or mistral_install_tagged|changed
+
```
armab: Please include missing `st2mistral` tag

roles/st2mistral/handlers/main.yml
```diff
@@ -1,3 +1,11 @@
+- name: enable mistral upgrade
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
+    - /etc/mistral/mistral-db-manage.populate.ansible.has.run
```
armab: Previous version looks good to me, just adding `skip_ansible_lint` tag to the task will make the CI happy. Ansible linter is being dumb here.

roles/st2mistral/tasks/main.yml
```diff
@@ -31,6 +33,11 @@
   when:
     - st2mistral_version != "latest"
     - st2mistral_version != "present"
+  notify: enable mistral upgrade
+  tags: st2mistral
+
+- name: Enable mistral upgrade
+  meta: flush_handlers
```
armab: Flushing handlers in the middle is usually a bad/undesired practice and should be made in rare cases as possible.
I understand this comes from the `enable mistral upgrade` handler. We can skip the Ansible linter for that task.

roles/st2mistral/tasks/main.yml
```diff
@@ -31,8 +33,19 @@
   when:
     - st2mistral_version != "latest"
     - st2mistral_version != "present"
+  register: mistral_install_tagged
   tags: st2mistral
 
+- name: Enable mistral db upgrade
+  file:
```
armab: One final thing is missing: `become: yes` to make sure we delete the lock files for sure

cognifloyd: doh!

cognifloyd:
fixed, rebased, and squashed.

PR #186, opened 2017-12-09
[Temp] El6: Pin postgres to fix "initdb" issue in postgresql94-9.4.15-3

humblearner:
Refer: https://github.com/StackStorm/st2-packages/pull/527

### DIFF
--- a/roles/postgresql/tasks/postgresql_yum6.yml
+++ b/roles/postgresql/tasks/postgresql_yum6.yml
@@ -28,7 +28,7 @@
     name: "{{ item }}"
     state: installed
   with_items:
-    # Temp pinning as "/var/lib/pgsql/9.4/data/pg_hba.conf does not exist on postgresql9494-9.4.15-2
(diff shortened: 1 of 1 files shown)


roles/postgresql/tasks/postgresql_yum6.yml
```diff
@@ -28,9 +28,15 @@
     name: "{{ item }}"
     state: installed
   with_items:
-    - postgresql94-server
-    - postgresql94-contrib
-    - postgresql94-devel
+    # Temp pinning as "/var/lib/pgsql/9.4/data/pg_hba.conf does not exist on postgresql9494-9.4.15-2
```
Mierdin: Just a little nitpick but the problem version was `postgresql9494-9.4.15-3`, and the reason for the failure was the `initdb` failure didn't create this file - it should still exist in a proper install.

Again, that context is in my PR which you referenced, so not a big deal, just might be a bit confusing if someone doesn't go hunt that context down.

PR #188, opened 2017-12-15
st2chatops: allow auth passwords with spaces in "st2 auth" command

kevin-vh:
Passwords with spaces were passed incorrectly because of missing quotes. This caused the executed command to fail as the space separated parts were each used as argument instead of as one argument.


armab:
Included in `v0.8.6`: https://galaxy.ansible.com/StackStorm/stackstorm/

PR #200, opened 2018-09-21
Multiple kitchen test suites default to same playbook

cognifloyd:
This makes sure that when any new test suites get added, they all use
the same playbook (stackstorm.yml) by default, no matter what the test
suite is named. If the suite needs a different playbook, the override
can be placed in test/integration/<suite name>/default.yml

### DIFF
--- a/test/integration/default.yml
+++ b/test/integration/default.yml
@@ -1,0 +1,1 @@
+../../stackstorm.yml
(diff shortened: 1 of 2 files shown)


cognifloyd:
Extracted from #162 as it is a safe incremental step towards running multiple test suites and doesn't depend on anything in that PR.

PR #198, opened 2018-07-10
Added st2workflowengine

nmaludy:
Closes #196
Added st2workflowengine to the list of services for 2.8.0 release

PR #201, opened 2018-09-22
Add new 2.9 service: timersengine

### LINKED ISSUE #196: Add new service st2timersengine
https://github.com/StackStorm/st2docs/pull/749

StackStorm core is adding a new service `st2timersengine`. Need to add this to the service list: https://github.com/StackStorm/ansible-st2/blob/master/roles/st2/vars/main.yml#L5-L14

This way the service is enabled and started.

cognifloyd:
Add new 2.9 service: timersengine

Resolves #196

### DIFF
--- a/roles/st2/vars/main.yml
+++ b/roles/st2/vars/main.yml
@@ -1,7 +1,7 @@
 # Default st2 immutable vars
 ---
 # List of available `st2` services:
-# https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl#L5
+# https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl#L17
 st2_services:
   - st2actionrunner
   - st2garbagecollector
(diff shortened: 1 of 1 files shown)


cognifloyd:
I don't understand the test results. On systemd systems this passes, but the idempotence check fails on sysv/upstart because the st2workflowengine and st2timersengine both report changed every time they're requested to start.

cognifloyd:
I'll drop the rest for now.

I'm surprised upgrading isn't supported as this is in the readme:
> st2_version | latest | StackStorm version to install. `present` to install available package, **`latest` to get automatic updates**, or pin it to numeric version like `2.2.0`.

So, installing st2 via ansible supports automatic updates but not upgrades?

cognifloyd:
dropped upgrade logic and rebased on #200.

armab:
That's a good and a valid question.

For Ansible I think we tend to follow slightly more simplified approach comparing to traditional Chef or Puppet ways.
It's a tradeoff we have to do balancing between complexity vs able to support it.
Apart of this migration scenarios there were more tough corner cases. If we add a few migrations today and don't support future migrations that would be not consistent.

I'll look at this PR in more detail later post 2.9 release. In a meantime, can you please create a new PR just for adding `st2timersengine` similar to https://github.com/StackStorm/ansible-st2/pull/198?

cognifloyd:
I have a copy of the upgrade logic saved in a separate branch. This is now just the timersengine change.

PR #204, opened 2018-10-01
Add pip role

cognifloyd:
This embeds the pip role from https://github.com/bobbyrenwick/ansible-pip

The role is licensed under Apache 2.0.

A follow up PR will add the role to the playbook wherever necessary.
It is important to make sure pip is installed on the managed host for
these reasons:

1. If an ansible module has any python dependencies, those need to be
   installed with whatever python ansible is using on the managed host.
2. Installing those dependencies as system packages, instead of using
   pip to install them, will often install very outdated packages.
3. Pip is a very basic package that most systems need anyway.
4. Pretty pretty please with a cherry on top.

Background:
I want to use the mongodb_* ansible modules in a follow up PR, but they
require the pymongo python module. The containers containers used in
Travis do not have pip by default, so to get a new enough verision of
pymongo installed, I need to install pip first.


armab:
Cool, worth updating #162 then and so we also get rid of requirements.txt dependency.

cognifloyd:
I already dropped requirements in that PR. This was my attempt to add a role to install pip more cleanly than the couple of tasks I replaced the requirements with in there. I closed #162 anyway. #205 contains the next step towards adding mongo auth.

armab:
Alright!

Issue #190, opened 2018-03-28
Ansible 2.4+ deprecation warnings


armab:
Fix Ansible deprecation warnings starting to appear since Ansible `2.4`: 
```
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions.
 This feature will be removed in a future release. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is
 discouraged. The module documentation details page may explain more about this
 rationale.. This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

cognifloyd:
`import_tasks` and `include_tasks` require ansible 2.4 and up.
`include` was set to be deprecated in [ansible 2.8](https://github.com/ansible/ansible/blob/8799038a895ed0ff230e2aa0629c4b83e8f15cf1/lib/ansible/modules/utilities/logic/_include.py#L22), but that has been [delayed](https://github.com/ansible/ansible/commit/4dcc5588f695dba5a04dfc4ebdbb73818a255f51#diff-fa05a7dabb0c43b34ba1eeb3464bd2a2):
> defering deprecation until we are comfortable that the new actions
> cover most of the ground users need.

If we're ready to require ansible 2.4 minimum then we can get rid of the deprecation notices. I would love to require ansible 2.4 as I already require it for the playbooks I write. But, I don't know if that would be a generally acceptable requirement yet.

armab:
@cognifloyd That's a perfectly valid point, agree with you :+1: 

So yeah, Ansible 2.5 is just out.
Considering that currently we set minimum `2.2` requirement in roles, probably pumping requirements to `2.4` would be too much an aggressive move.

Will wait for a couple of Ansible releases for things to settle down and users move to new versions, before making this change.

cognifloyd:
:tada: Ansible 2.6 is here. https://groups.google.com/forum/#!topic/ansible-announce/JI1Zju5x3TA

It's got lots of bugfixes for include* and import*.

Current ansible [support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html) for 2.2 and above has not been updated to account for the release of 2.6:

Ansible Release | Latest Version | Status
-- | -- | --
[...]

PR #210, opened 2018-10-23
Bump Ansible version to 2.3

arhue:
https://github.com/StackStorm/ansible-st2/pull/208 bumps Ansible version to 2.3.

@armab There you go!

PR #211, opened 2018-10-26
Changed use of Jinja filters as tests

LindsayHill:
The use of `when mistral_install_latest|changed` is now [deprecated](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.5.html#jinja-tests-used-as-filters).

This should be `when mistral_install_latest is changed`

Looks like this will require a minimum Ansible version of 2.5, but this is also the oldest version supported by upstream. We should plan to change our [minimum Ansible version to 2.5](https://github.com/StackStorm/ansible-st2/issues/190).

### DIFF
--- a/roles/st2mistral/tasks/main.yml
+++ b/roles/st2mistral/tasks/main.yml
@@ -45,7 +45,7 @@
   with_items:
     - /etc/mistral/mistral-db-manage.upgrade.head.ansible.has.run
     - /etc/mistral/mistral-db-manage.populate.ansible.has.run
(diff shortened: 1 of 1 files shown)


armab:
> Using Ansible-provided jinja tests as filters will be removed in Ansible 2.9.

Hm, I think it's too early for us to include change like this and worth waiting a little bit more.

In https://github.com/StackStorm/ansible-st2/pull/210 we just bumped min Ansible version to `v2.3` which is OK, but setting up minimum `v2.5` is too much.
Needs more time.

armab:
I've tested this change with older version and @cognifloyd is right, it works OK with Ansible before `v2.5.0`

PR #216, opened 2018-11-12
Mongo Auth | Intall pymongo via pip (vs via sys pkgs)

cognifloyd:
This PR is preparation for enabling mongo auth via the ansible playbook.
These tasks will not be activated until all required tasks are committed.

This installs pip via system package, and then pymongo via pip.

pymongo is needed for mongodb_user module to enable mongo auth.
pymongo 3.2+ is required to work with mongo 3.2+
system packages are too old on all but Ubuntu 16:
 - 2.5.2 on EL 6 and EL 7
 - 2.6.3 on Ubuntu 14 Trusty
 - 3.2 on Ubuntu 16 Xenial

### DIFF
--- a/roles/StackStorm.mongodb/tasks/mongodb_auth.yml
+++ b/roles/StackStorm.mongodb/tasks/mongodb_auth.yml
@@ -7,8 +7,9 @@
   tags: [databases, mongodb]
 
 - name: Install pymongo (for the mongodb_user module)
-  # pymongo 3.2+ is required to work with mongo 3.2+. system packages are too old.
+  # Use pip because system packages are too old for adequate mongodb support.
+  # https://docs.mongodb.com/ecosystem/drivers/driver-compatibility-reference/#python-driver-compatibility
   become: yes
   pip:
-    name: "pymongo=={{ mongodb_major_minor_version }}"
+    name: "pymongo>=3.10.1,<4.0.0"
   tags: [databases, mongodb]
--- a/roles/StackStorm.mongodb/tasks/mongodb_auth.yml
+++ b/roles/StackStorm.mongodb/tasks/mongodb_auth.yml
@@ -1,6 +1,14 @@
 ---
-- name: Ensure pymongo is installed for the mongodb_user ansible module
+- name: Install pip (for the installation of pymongo)
   become: yes
   package:
-    name: python-pymongo
+    name: python-pip
     state: present
+  tags: [databases, mongodb]
+
+- name: Install pymongo (for the mongodb_user module)
+  # pymongo 3.2+ is required to work with mongo 3.2+. system packages are too old.
+  become: yes
+  pip:
+    name: "pymongo=={{ mongodb_major_minor_version }}"
+  tags: [databases, mongodb]


roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -1,6 +1,12 @@
...
   package:
-    name: python-pymongo
+    name: "python{% if ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '7' %}2{% endif %}-pip"
     state: present
+
+- name: Install pymongo (for the mongodb_user module)
+  # pymongo 3.2+ is required to work with mongo 3.2+. system packages are too old.
+  become: yes
+  pip:
+    name: pymongo
```
armab: Let's please pin required `pymongo` via version specifiers

Can reuse
https://github.com/StackStorm/ansible-st2/blob/5eec6ac26b647cd9e89c1220a25e06ea14782f60/roles/mongodb/vars/main.yml#L2

roles/mongodb/tasks/mongodb_auth.yml
```diff
@@ -1,6 +1,12 @@
 ---
-- name: Ensure pymongo is installed for the mongodb_user ansible module
+- name: Install pip (for the installation of pymongo)
   become: yes
   package:
-    name: python-pymongo
+    name: "python{% if ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '7' %}2{% endif %}-pip"
```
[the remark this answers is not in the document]
cognifloyd: Ah cool. I did a `yum search` / `yum info` for it on a CentOS 7 machine. I didn't realize there could be aliases (is that what they're called) that don't show up in yum search/info. I'll gladly drop that special-case :+1:

cognifloyd:
Removed special case for EL7 and pinned version.
Tests here (with the rest of the auth tasks): https://travis-ci.org/theatro/ansible-st2/builds/454176993

cognifloyd:
Travis tests pass except for one false negative a chatops EADDRINUSE error.

cognifloyd:
@armab done!

cognifloyd:
Checks are green. Please review and merge.

roles/StackStorm.mongodb/tasks/mongodb_auth.yml
```diff
@@ -1,6 +1,14 @@
...
-    name: python-pymongo
+    name: python-pip
     state: present
+  tags: [databases, mongodb]
+
+- name: Install pymongo (for the mongodb_user module)
+  # pymongo 3.2+ is required to work with mongo 3.2+. system packages are too old.
+  become: yes
+  pip:
+    name: "pymongo=={{ mongodb_major_minor_version }}"
```
armab: Latest `pymongo` is `3.10.1` as for  `Jan 8, 2020 ` https://pypi.org/project/pymongo/

What will happen if `mongodb_major_minor_version` would be `4.0` if needed for Bionic #253 and EL8?
 
I don't think pymongo pip version correlates with `mongodb_major_minor_version`.

cognifloyd: Ah. I guess my sample size was too small when I looked that up. For versions 3.2-3.6, that assumption held.
Here's the chart showing supported versions: https://docs.mongodb.com/ecosystem/drivers/driver-compatibility-reference/#python-driver-compatibility
Now to find an elegant way to record those requirements.

armab: I think it's possible to use one of the latest versions pinned with safe interval.
My understanding that this should work for all older MongoDB versions.

armab: `pymongo>=3.10.1,<4.0.0`

cognifloyd: K. the latest commit pins the version as suggested.

PR #217, opened 2018-11-30
Migrating non-production GPG keys

bigmstone:
We are following through with migrating our PackageCloud repositories to
new GPG keys. This PR adds the new key IDs for all non-production
repositories as importable variables. Import is based on the selected
repository for both Enterprise and Open Source.


roles/bwc/tasks/bwc_repos_apt.yml
```diff
@@ -12,12 +12,59 @@
...
+  changed_when: no
+  command: cat "/etc/packagecloud/StackStorm_{{ bwc_repo }}_gpgkey_url.txt"
+  register: _bwc_gpgkey_url
+
+- name: Set bwc_gpgkey_url variable
+  no_log: yes
+  set_fact:
+    bwc_gpgkey_url: "{{ _bwc_gpgkey_url.stdout }}"
+
+  # This is a nasty hack necessary because of how AWS Redirects are interfearing
```
warrenvw: sp.  interfering

blag: Good catch. I pushed a fix.

PR #218, opened 2018-12-11
Rollback release version from 3.0 to 2.10

m4dcoder:
Release version has been rollbacked from 3.0 to 2.10 due to project management decision.


m4dcoder:
@armab thanks for pointing out the el7 issue.  it's fixed now and el7 packages are in unstable.  so let's see if rerun will succeed here.

PR #221, opened 2018-12-19
New linting rules

LindsayHill:
`ansible-lint` 4.0.0 introduces many new default rules. These trip us up in many places. 

The only new default rule I have not fixed is the 120-char line length rule. For now I have just excluded that rule from checks.

I have also fixed a small issue with our CircleCi config, where it was running ansible-lint against non-existent `*.yaml` files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,4 +18,4 @@ jobs:
       - run:
           name: Ansible-lint check
           command: |
-            ansible-lint -v roles/*/*/*.yaml roles/*/*/*.yml stackstorm.yml
+            ansible-lint -x 204 -v roles/*/*/*.yml stackstorm.yml
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,4 +18,4 @@ jobs:
       - run:
           name: Ansible-lint check
           command: |
-            ansible-lint -x 204 -v roles/*/*/*.yml stackstorm.yml
+            ansible-lint -x 204 -v roles/*/*/*.yaml roles/*/*/*.yml stackstorm.yml
--- a/roles/bwc/tasks/bwc_repos_setup.yml
+++ b/roles/bwc/tasks/bwc_repos_setup.yml
@@ -63,4 +63,4 @@
     - ewc
     - enterprise
   register: bwc_repo_added
-  when: bwc_read_token != ''
+  when: bwc_read_token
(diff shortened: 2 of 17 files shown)


roles/st2chatops/tasks/main.yml
```diff
@@ -82,7 +82,7 @@
 
 - name: Generate authentication token
   command: st2 auth "{{ st2_auth_username }}" -p "{{ st2_auth_password }}" -t
-  when: task_apikey_not_exists is succeeded and task_user_st2_api_key.changed == false
+  when: task_apikey_not_exists is succeeded and task_user_st2_api_key.changed
```
cognifloyd:
```suggestion
  when: task_apikey_not_exists is succeeded and task_user_st2_api_key is not changed
```
Using `.changed` like this is not the normal way to test this. We should use actual tests. Same with other tasks below where `.changed` should be replaced with ` is changed` or ` is not changed` (depending on the test).

.circleci/config.yml
```diff
@@ -18,4 +18,4 @@ jobs:
       - run:
           name: Ansible-lint check
           command: |
-            ansible-lint -v roles/*/*/*.yaml roles/*/*/*.yml stackstorm.yml
+            ansible-lint -x 204 -v roles/*/*/*.yml stackstorm.yml
```
armab:
```
WARNING: Couldn't open roles/*/*/*.yaml - No such file or directory
```

Some people name files `.yaml`, some `.yml`. Both formats are perfectly valid and working in Ansible.
And while I renamed `yaml` -> `yml` several times, there is a chance that someone will commit new `yaml` files in future.

So by adding this additional `yaml` rule I want to make sure we're not missing any files here.

It's not a bug, but a feature :)

roles/bwc/tasks/bwc_repos_setup.yml
```diff
@@ -63,4 +63,4 @@
     - ewc
     - enterprise
   register: bwc_repo_added
-  when: bwc_read_token != ''
+  when: bwc_read_token
```
armab: I guess it's different comparing to checking for empty string.
If linter is not happy with that, maybe `bwc_read_token | length > 0` would work?

roles/st2chatops/tasks/main.yml
armab: Good catch!
That's most probably the reason why CI smoketests are not happy in this PR.

LindsayHill:
Thanks for the input both of you. Have merged your suggestions. Now just have to wait for Travis to do its thing. Assuming all is well, I will then merge

PR #231, opened 2019-02-19
Fix ansible-lint CI failures

armab:
`ansible-lint` recently introduced a few more checks which made our CI to fail.
Update the code to rely on linter updates.


roles/StackStorm.ewc/tasks/ewc_repos_cleanup_yum.yml
```diff
@@ -1,8 +1,7 @@
 ---
-
 - name: Cleanup repo list file from disk
   become: yes
-  yum_resository:
+  yum_repository:
```
armab: That made my coding day, @lakshmi-kannan's 2 yrs old typo ;)

PR #233, opened 2019-04-05
Fix an issue with st2api not being restarted after installing bwc-enterprise package, fix services which are restarted when configuring ldap backend in st2.conf

Kami:
This pull request fixes two issue:

1. Make sure ``st2api`` is restarted after ``bwc-enterprise`` package is installed so RBAC indeed enabled and working (this is needed in StackStorm >= 3.0.0)
2. Restart ``st2auth`` when configuring LDAP backend in st2.conf - previously incorrect services were restarted

### DIFF
--- a/roles/StackStorm.ewc/tasks/main.yml
+++ b/roles/StackStorm.ewc/tasks/main.yml
@@ -24,6 +24,8 @@
     - ewc
     - st2 enterprise
     - skip_ansible_lint
+  notify:
+    - restart st2api
 
 - name: Install present bwc-enterprise package, no auto-update
   become: yes
@@ -38,6 +40,8 @@
   tags:
     - ewc
     - st2 enterprise
+  notify:
+    - restart st2api
 
 - name: Install pinned bwc-enterprise package
   become: yes
@@ -54,6 +58,8 @@
   tags:
     - ewc
     - st2 enterprise
+  notify:
+    - restart st2api
 
 - name: Setup RBAC and setup roles and assignments if ewc_rbac is defined
   import_tasks: rbac.yml
--- a/roles/StackStorm.ewc/tasks/main.yml
+++ b/roles/StackStorm.ewc/tasks/main.yml
@@ -26,6 +26,7 @@
     - skip_ansible_lint
   notify:
     - restart st2api
+    - restart st2auth
 
 - name: Install present bwc-enterprise package, no auto-update
   become: yes
@@ -42,6 +43,7 @@
     - st2 enterprise
   notify:
     - restart st2api
+    - restart st2auth
 
 - name: Install pinned bwc-enterprise package
   become: yes
@@ -60,6 +62,7 @@
     - st2 enterprise
   notify:
     - restart st2api
+    - restart st2auth
 
 - name: Setup RBAC and setup roles and assignments if ewc_rbac is defined
   import_tasks: rbac.yml
--- a/roles/StackStorm.ewc/tasks/rbac.yml
+++ b/roles/StackStorm.ewc/tasks/rbac.yml
@@ -55,6 +55,7 @@
     backup: yes
   notify:
     - restart st2api
+    - reload ewc_rbac
 
 - name: Configure RBAC backend in st2 configuration
   become: yes
@@ -66,3 +67,4 @@
     backup: yes
   notify:
     - restart st2api
+    - reload ewc_rbac
--- a/roles/StackStorm.ewc/tasks/rbac.yml
+++ b/roles/StackStorm.ewc/tasks/rbac.yml
@@ -57,3 +57,16 @@
     - restart st2
     - reload ewc_rbac
     - restart st2api
+
+- name: Configure RBAC backend in st2 configuration
+  become: yes
+  ini_file:
+    dest: /etc/st2/st2.conf
+    section: rbac
+    option: backend
+    value: enterprise
+    backup: yes
+  notify:
+    - restart st2
+    - reload ewc_rbac
+    - restart st2api
--- a/roles/StackStorm.ewc/tasks/rbac.yml
+++ b/roles/StackStorm.ewc/tasks/rbac.yml
@@ -54,8 +54,6 @@
     value: True
     backup: yes
   notify:
-    - restart st2
-    - reload ewc_rbac
     - restart st2api
 
 - name: Configure RBAC backend in st2 configuration
@@ -67,6 +65,4 @@
     value: enterprise
     backup: yes
   notify:
-    - restart st2
-    - reload ewc_rbac
     - restart st2api
--- a/roles/StackStorm.ewc_smoketests/tasks/main.yml
+++ b/roles/StackStorm.ewc_smoketests/tasks/main.yml
@@ -65,7 +65,7 @@
     - ewc-smoke-tests
 
 - name: Test some other action that "{{ ewc_smoke_tests_user }}" cannot run
-  command: st2 run core.http url="https://google.com"
+  command: st2 run core.http url="https://www.google.com"
   environment:
     ST2_AUTH_TOKEN: "{{ st2_token_smoke_tests_user.stdout }}"
   ignore_errors: yes
--- a/roles/StackStorm.ewc/tasks/ldap.yml
+++ b/roles/StackStorm.ewc/tasks/ldap.yml
@@ -12,8 +12,7 @@
   # Don't even setup LDAP if backend_kwargs is not defined
   when: ewc_ldap.backend_kwargs is defined and ewc_ldap.backend_kwargs|length > 0
   notify:
-    - restart st2api
-    - restart st2stream
+    - restart st2auth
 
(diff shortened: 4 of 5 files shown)


Kami:
And second commit (a999b4287f78617a6ffc53b2db0857d798023ec7) is fixed for RBAC in v3.0 dev where a new st2.conf config option was added.

This change is backward compatible and also works with older versions (it's simply ignored in older versions).

It looks like tests are still failing though, so I need to dig in.

Kami:
Digging in a bit - it looks like st2api is last restarted here (https://travis-ci.org/StackStorm/ansible-st2/jobs/516373256#L1536), before installing bec-enterprise package and before RBAC is enabled and backend configured here (https://travis-ci.org/StackStorm/ansible-st2/jobs/516373256#L1733).

So it looks like the issue is related to st2api not being restarted after installing bwc-enterprise package and configuring RBAC.

roles/StackStorm.ewc_smoketests/tasks/main.yml
```diff
@@ -65,7 +65,7 @@
     - ewc-smoke-tests
 
 - name: Test some other action that "{{ ewc_smoke_tests_user }}" cannot run
-  command: st2 run core.http url="https://google.com"
+  command: st2 run core.http url="https://www.google.com"
```
armab: This is not an initial issue here, see ignore_errors a few lines below.

roles/StackStorm.ewc/tasks/rbac.yml
```diff
@@ -54,6 +54,15 @@
     value: True
     backup: yes
   notify:
-    - restart st2
-    - reload ewc_rbac
+    - restart st2api
+
+- name: Configure RBAC backend in st2 configuration
```
armab: This should be removed as it’s part of the deb/rpm packaging scriplets.
Duplicating steps doesn’t help but instead makes harder to find issues in package installation results.

Kami:
It looks like everything is green now - https://travis-ci.org/StackStorm/ansible-st2/builds/516385105

I pushed a change so st2api is restarted when installing / upgrading ``bwc-enterprise`` package. That is something which indeed always needs to happen (even before my RBAC changes), otherwise RBAC won't work.

I checked some old builds (https://travis-ci.org/StackStorm/ansible-st2/jobs/515462624#L1925) and it looks like tests mostly passed due to "luck" (aka if st2 restart was scheduled after installing bwc-enterprise package and enabling RBAC) or they didn't run at all and were skipped (https://travis-ci.org/StackStorm/ansible-st2/jobs/498284336#L1799).

To be on the "safe side", we could also update "Configure RBAC backend in st2 configuration" task to always notify ``restart st2api``, even if there are no changes.

/cc @arma @nmaludy

roles/StackStorm.ewc/tasks/rbac.yml
Kami: I think it's still better / safer to leave it here.

It doesn't hurt and it won't trigger a notification anyway if the option is already present.

Kami:
To add some more context to the original failed build @armab linked to me https://travis-ci.org/StackStorm/ansible-st2/jobs/515972711#L1700.

As you can see at log line 1700 (https://travis-ci.org/StackStorm/ansible-st2/jobs/515972711#L1700), ``bwc-enterprise`` package is installed. After that, up to the line 1936 where smoke tests run (https://travis-ci.org/StackStorm/ansible-st2/jobs/515972711#L1936) ``st2api`` is never restarted.

As mentioned above, this would never work - to enable RBAC after installing ``bwc-enterprise`` package (which automatically configures ``rbac`` section in ``st2.conf``), ``st2api`` needs to be restarted and as per log file, this doesn't appear to be happening.

Last time st2api is restarted is at line 1532 (https://travis-ci.org/StackStorm/ansible-st2/jobs/515972711#L1532) which is **before** bwc-enterprise is install.

So I think my original conclusion from the comment above is still accurate.

On a related note of ``bwc-enterprise`` package install and upgrade automatically restarting ``st2api`` service - both @armab and I agreed earlier this week that this is not a good idea - we should give user control when and how to restart services during the upgrade and not do it automatically.

roles/StackStorm.ewc/tasks/rbac.yml
```diff
@@ -54,6 +54,15 @@
...
+  become: yes
+  ini_file:
+    dest: /etc/st2/st2.conf
+    section: rbac
+    option: backend
+    value: enterprise
+    backup: yes
+  notify:
     - restart st2api
```
armab: Same as in previous block: 
```suggestion
    - restart st2api
    - reload ewc_rbac
```

roles/StackStorm.ewc/tasks/main.yml
```diff
@@ -24,6 +24,8 @@
     - ewc
     - st2 enterprise
     - skip_ansible_lint
+  notify:
+    - restart st2api
```
armab: Because `bwc-enterprise` package also includes LDAP auth backend, are you sure `st2api` restart is sufficient or `st2auth restart` is needed too?

Kami: Good catch - we should also restart st2auth.

This should happen after configuring auth backed in st2.cond though

armab: That already happens https://github.com/StackStorm/ansible-st2/blob/master/roles/StackStorm.ewc/tasks/ldap.yml, however edge case is when the installation wasn't managed by Ansible before or user changed st2.conf by hand/uninstalled/installed/upgraded enterprise package.

armab: https://github.com/StackStorm/ansible-st2/blob/ae27046e29ed9f890a3b138512fbd3f66dcb30b2/roles/StackStorm.ewc/tasks/ldap.yml#L13-L16

`st2stream` restart might be needed, too?

Kami: nope, it should only affect st2auth. st2api and st2stream only work with auth token / API key and are not aware of any auth backed related stuff.

armab:
> To be on the "safe side", we could also update "Configure RBAC backend in st2 configuration" task to always notify restart st2api, even if there are no changes.

This goes against the configuration management principles of idempotence and best practices, when the action should be done only when it's required.
If we do it differently, we'll restart important services on live production servers on every Ansible run resulting in potential loss of traffic and other undesired consequences which is exactly what we try to avoid by adding idempotence check in `ansible-st2` CI.

Kami:
> This goes against the configuration management principles of idempotence and best practices, when the action should be done only when it's really required.

I don't agree about the idempotence definition part - idempotent operation means that it can be executed multiple times and will always yield the same result - which is indeed the case with setting the config option operation.

And I'm also very skeptical about this part in general:

> when the action should be done only when it's really required.

I used puppet and Chef on large fleet of servers for a long time and usually the main problem which wasted everyone's time debugging things was Puppet trying to be too smart when to apply something (aka if something has changed or not or similar).

In theory, it always sounds like a nice thing, but in practice it resulted in tons of hard to track down and troubleshoot issues and tons of wasted time.

In the end we ended up rewriting most recipes (where possible and applicable) to be idempotent and simply always perform the operation. It resulted in recipes which were much easier to reason about and debug.

Anyway, that's a bit off topic :)

roles/StackStorm.ewc/tasks/main.yml
Kami: And after more thought and looking at the code - I don't think this need or should happen here.

LDAP backend is not used and enabled by default when installing ``bwc-enterprise`` (aka that's different than ``st2-rbac-backend`` pacakge).

So st2api and st2auth needing to be restarted is purely depended on the ldap playbook - when ldap is configured and configuration changes.

Kami:
To wrap those changes up since there is a lot of noise and also some slightly off topic discussion here.

This pull request fixes a bug which was related to st2api not being restarted after ``bwc-enterprise`` package was installed / upgrade. This is now needed because ``st2-rbac-backend`` changes automatically configures ``st2.conf`` and enabled RBAC (but doesn't restart ``st2api``).

Previously restart was triggered after ansible modified st2.conf when configuring RBAC.

Second issue is unrelated to this change, but we randomly noticed it - 1a144fb. We should restart st2auth after configuring and enabling enterprise backend. Those changes only affect auth - st2api and st2stream work directly with auth tokens and api keys - they don't know or care anything about the auth backend itself.

roles/StackStorm.ewc/tasks/main.yml
armab: See previous comment: 
> however edge case is when the installation wasn't managed by Ansible before or user changed st2.conf by hand/uninstalled/installed/upgraded enterprise package

and so `st2auth` restart should happen here as well.

This will cover case when user already had settings for LDAP configured in `st2.conf` (which doesn't trigger st2auth restart), he then upgraded the enterprise package to new version. In such scenario old py code will be used instead of new code of ldap package.

Issue #156, opened 2017-07-27
when setting st2.conf variables, double quotes are converted to single quotes


johnarnold:
Example in ansible:
```
- name: auth | Configure ldap backend
  become: yes
  ini_file:
    dest: /etc/st2/st2.conf
    section: auth
    option: backend_kwargs
    value: '{ "ldap_uri": "ldaps://ad.domain.com:3269", "use_tls": false, "bind_dn": "user@domain.com", "bind_pw": "{{ _user_pass }}", "user": {"base_dn": "CN=Users,DC=domain,DC=com", "search_filter": "(&(objectClass=user)(sAMAccountName={username})(|(memberOf=CN=group1,OU=SecurityGroups,DC=domain,DC=com)(memberOf=CN=group2,OU=SecurityGroups,DC=domain,DC=com)))", "scope": "subtree"}}'
    backup: yes
```
Results in:

```
backend_kwargs = {'use_tls': false, 'bind_pw': 'a_string', 'ldap_uri': 'ldaps://ad.domain.com:3269', 
...
```

I believe ansible is interpretting everything in the value: as json, and converting it into ?yaml? internally.  Possibly related:  https://www.reddit.com/r/ansible/comments/4m1njk/whats_the_proper_way_to_escape_double_quotes_in/

I'm not sure what the best "fix" is, but a workaround is to just put a space at the beginning of your value, like:

value: ' { ...

armab:
We have related logic in `bwc` role where we set the `backend_kwargs` https://github.com/StackStorm/ansible-st2/blob/d7ca9a9d6aecc3f0a5de477ba0923c752faf4cc4/roles/bwc/tasks/ldap.yml#L27
It works as expected, see the configuration example with `bwc_ldap:` https://docs.stackstorm.com/install/ansible.html#bwc-st2-enterprise

Maybe `var | string` is the magic that did the trick for us.

armab:
Closing, since it's not a problem of `ansible-st2` codebase.

lingfish:
I know this is closed, but was an answer really provided?  @armab are you suggesting we have to hack this in ourselves, in order to configure "auth other than BWC's one"?

armab:
https://github.com/StackStorm/ansible-st2/pull/234 should help to manage it in the following way from `st2` role, `st2_config` var:

```yaml
---
- name: Install st2
  hosts: all
  roles:
    - role: StackStorm.st2
      vars:
        st2_config:
          auth:
[...]

PR #234, opened 2019-04-06
Ensure st2.conf values are not converted to particular structure

armab:
Closes #238 

Covers edge case when user wants to set in st2.conf json-formatted value which might be converted to particular structure during the jinja/ansible pipeline.

See https://github.com/StackStorm/ansible-st2/issues/156


Kami:
LGTM.

How hard would it be to add tests for it?

armab:
That would be hard and bring not that sufficient value comparing to increasing the build timing/complexity.
Ansible plays we have are designed to be simple and have only smoke tests instead of checking all the permutations of possible configurations. It's the best effort we can do and is good enough, considering ongoing development/maintenance of many other projects.

PR #235, opened 2019-04-06
Fix 'ewc_license' conditional check in stackstorm.yml example play

armab:
Closes #237 
Under some circumstances `ewc_license` could be `None` in `stackstorm.yml` Ansible play

Fixes:
>fatal: [ubuntu16]: FAILED! => {"msg": "The conditional check 'ewc_license is defined and ewc_license | length > 0' failed. The error was: Unexpected templating type error occurred on ({% if ewc_license is defined and ewc_license | length > 0 %} True {% else %} False {% endif %}): object of type 'NoneType' has no len()\n\n"}


lingfish:
Heh, I just did a PR for this as well.  I believe that, if you guys have chosen to always set a default in vars, when the defined check is redundant.

PR #239, opened 2019-04-08
st2chatops also may need proxies

lingfish:
Otherwise it can't connect to RTM on Slack, for example.

### DIFF
--- a/roles/StackStorm.st2/tasks/proxy.yml
+++ b/roles/StackStorm.st2/tasks/proxy.yml
@@ -10,7 +10,7 @@
     # NB: Empty ENV var cast to 'None' string in Ansible
     state: "{{ 'present' if ansible_facts.env.get(item.1, 'None') != 'None' else 'absent' }}"
   vars:
-    _services: [st2api, st2actionrunner]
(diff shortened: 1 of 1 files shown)


roles/StackStorm.st2/tasks/proxy.yml
```diff
@@ -10,7 +10,7 @@
     # NB: Empty ENV var cast to 'None' string in Ansible
     state: "{{ 'present' if ansible_facts.env.get(item.1, 'None') != 'None' else 'absent' }}"
   vars:
-    _services: [st2api, st2actionrunner]
+    _services: [st2api, st2actionrunner, st2chatops]
```
armab: After looking closer, this addition should belong to `st2chatops` Ansible role as it could be installed separately.

lingfish: Ok, how can we do this when this one has already been merged?

armab: Just open another PR that adds to `st2chatops` and removes from `st2`

PR #244, opened 2019-06-06
Explicit mark variable as bool

LindsayHill:
Resolve a couple more deprecation warnings, e.g.

```
TASK [StackStorm.st2 : user | Disable requiretty] *************************************************************************************************************************
[DEPRECATION WARNING]: evaluating st2_system_user_in_sudoers as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the
future. Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.
```

### DIFF
--- a/roles/StackStorm.rabbitmq/tasks/main.yml
+++ b/roles/StackStorm.rabbitmq/tasks/main.yml
@@ -29,5 +29,5 @@
     new_only: no
     # the EL6 RPM for rabbitmq-server does not place the rabbitmq-plugins binary in the search path (other distro's packages do, including EL7). Prefix adds it to the binary search path.
     prefix: "{{ rabbitmq_on_el6 | ternary(rabbitmq_el6_prefix, omit) }}"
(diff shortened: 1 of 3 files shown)


armab:
I also found this Warning/behavior strange from Ansible side. At least warning shouldn't show if var is already bool.

armab:
Here we go, it's already pretty hot discussed indeed: https://github.com/ansible/ansible/issues/53428#issuecomment-497380458
and improved a couple of days ago: https://github.com/ansible/ansible/pull/57190

We'll probably see improvements in next Ansible version. Or hopefully they'll backport it to latest stable.

LindsayHill:
Yeah, it's pretty annoying behavior for me where that variable is a `bool`. I understand the warning in situations where it's being evaluated as truthy, but for these cases it's pretty petty.

PR #230, opened 2019-02-18
added ability to install ST2 packs

guymatz:
set st2_packs to have them installed.  This should be a module, but this is good enough for now

PR #250, opened 2019-08-30
Support Pack Install as a variable in stackstorm.st2 (py2)

punkrokk:
Closes #230

After looking at finishing https://github.com/StackStorm/ansible-st2/pull/230, it became clear that turning this into a role that supported py2 and py3 packs was desirable. That then required a task to install py3. I've modularized it so it's easy to remove when ST2 goes py3 only, and to only install if desired.


roles/StackStorm.st2/defaults/main.yml
```diff
@@ -34,3 +34,5 @@ st2_auth_username: testu
 st2_auth_password: testp
 # Save credentials in ~/.st2/config file
 st2_save_credentials: yes
+# ST2 packs to be installed (list)
+st2_packs: [azure, "algosec --python3"]
```
armab: Per https://github.com/StackStorm/ansible-st2/pull/230#issuecomment-491920683 we want `st2` core pack to be installed by default

roles/Stackstorm.post_install/tasks/packs.yml
```diff
@@ -0,0 +1,34 @@
+---
+- name: get authentication token
+  command: st2 auth "{{ st2_auth_username }}" -p "{{ st2_auth_password }}" -t
+  register: st2_token
```
armab: Having st2 pack install functionality in `st2` role comes with a bonus that we don't need to generate a token as additional step.

punkrokk: yea makes sense

roles/StackStorm.st2/defaults/main.yml
punkrokk: I am pulling this out (py3)

roles/Stackstorm.post_install/tasks/packs.yml
punkrokk: I actually don't see where this is generated before I need it.

armab: See https://github.com/StackStorm/ansible-st2/blob/684d0c3052b8fc3503baec06c9eaed3fd32e9d6b/roles/StackStorm.st2/tasks/auth.yml#L49-L68

`$HOME/.st2/config` file contains username/password pair meaning no further auth is required from st2client side.

punkrokk: I see it. I think I didn’t setup as root. How do you feel about ignore EL6/ Centos6 - and just not running it ? The failure for that is due to the map function to see if packs are installed not standard in EL6 @armab

armab: We'll deprecate EL6 in next 1-2 st2 releases so I'm good with that.

punkrokk:
@armab any idea why CentOS is able to pull creds per your suggestion but not Ubuntu?

armab:
@punkrokk Pay attention to build failure logs and warning around `flush_handlers`

punkrokk:
@armab seems like ubuntu builds don't have st2 logged in where I need to install packs. Here's proof: https://travis-ci.org/StackStorm/ansible-st2/jobs/580017011#L1311

It works fine locally in vagrant thought. Not sure how to troubleshoot this one.

roles/StackStorm.st2/tasks/main.yml
```diff
@@ -100,3 +100,15 @@
     state: started
   loop: "{{ st2_services }}"
   tags: st2
+
+# https://github.com/ansible/ansible/issues/41313#issuecomment-520891625
```
armab: :+1:
nice trick.
Can We also include a small clarifying comment besides of URL

roles/StackStorm.st2/tasks/flush_handlers.yml
```diff
@@ -0,0 +1,4 @@
+---
+
+- meta: flush_handlers
+  tags: st2_packs
```
armab: Missing new line at the end of file

roles/StackStorm.st2/defaults/main.yml
```diff
@@ -34,3 +34,5 @@ st2_auth_username: testu
 st2_auth_password: testp
 # Save credentials in ~/.st2/config file
 st2_save_credentials: yes
+# ST2 packs to be installed (list)
+st2_packs: [azure, jira]
```
armab:
```suggestion
st2_packs:
  -  st2
```

Per discussion in previous PR, we want to install only `st2` pack by default: https://github.com/StackStorm/ansible-st2/pull/230#issuecomment-491920683

roles/StackStorm.st2/tasks/main.yml
```diff
@@ -100,3 +100,15 @@
...
+- name: Flush handlers to prepare to run st2 there are packs to install
+  include_tasks: flush_handlers.yml
+  when: st2_packs is defined
+
+- name: Install ST2 add-on Packs
+  import_tasks: packs.yml
+  when: st2_packs is defined and
```
armab: `st2_packs` is always defined from default values.

Would be better to check if st2_packs list size is > 0

roles/StackStorm.st2/tasks/main.yml
```diff
@@ -100,3 +100,15 @@
...
+        ((ansible_distribution == "RedHat" and ansible_distribution_major_version == '7') or
```
armab: Including particular OSes in this check is error-prone as other OS versions could be added.
Please check if target OS is EL6 explicitly instead as that's only one that had issues:

See an existing code example: https://github.com/StackStorm/ansible-st2/blob/684d0c3052b8fc3503baec06c9eaed3fd32e9d6b/roles/StackStorm.st2/tasks/main.yml#L11

roles/StackStorm.st2/tasks/main.yml
```diff
@@ -100,3 +100,15 @@
...
+        (ansible_distribution == "Ubuntu"))
+  tags: st2, st2_packs, skip_centos_6
```
armab: Additionally, if target OS is `el6` can we print an error message indicating that `st2_packs` is not supported with EL6?

It could be done with `fail` Ansible task + `ignore_errors: yes`.

Vagrantfile
```diff
@@ -18,6 +18,7 @@ VIRTUAL_MACHINES = {
     :hostname => 'ansible-st2-centos7',
     :box => 'centos/7',
   },
+
```
armab:
```suggestion
```

roles/StackStorm.st2/tasks/main.yml
punkrokk: done

roles/StackStorm.st2/tasks/main.yml
punkrokk: ok refactored to `exclude` CentOS6

roles/StackStorm.st2/tasks/main.yml
punkrokk: done

roles/StackStorm.st2smoketests/tasks/main.yml
```diff
@@ -50,6 +50,14 @@
   tags:
     - smoke-tests
 
+- name: Verify st2 pack is installed
+  environment:
+    ST2_AUTH_TOKEN: "{{ st2_token.stdout }}"
+  command: "st2 pack install st2"
```
armab: `command` content looks like copy-pasted by mistake from the previous task.

armab: Also we'll need to remove previous task as pack install now happens in `st2` role

roles/StackStorm.st2smoketests/tasks/main.yml
```diff
@@ -41,12 +41,14 @@
...
-  become: yes
+- name: Get st2 pack info
   environment:
     ST2_AUTH_TOKEN: "{{ st2_token.stdout }}"
-  command: "st2 pack install st2"
+  command: st2 pack get st2 -j
   changed_when: no
+  check_mode: no
+  register: _smoke_st2_pack_installed
+  failed_when: _smoke_st2_pack_installed.rc != 0
```
armab: The Ansible task should already fail if it exits with non-zero code,
this check looks excessive.

roles/StackStorm.st2smoketests/tasks/main.yml
```diff
@@ -41,12 +41,14 @@
   tags:
     - smoke-tests
 
-- name: Install st2 pack from exchange
-  become: yes
+- name: Get st2 pack info
   environment:
     ST2_AUTH_TOKEN: "{{ st2_token.stdout }}"
-  command: "st2 pack install st2"
+  command: st2 pack get st2 -j
```
armab: :+1:

The only question: is there a good reason to output as json? 
```suggestion
  command: st2 pack get st2
```

roles/StackStorm.st2smoketests/tasks/main.yml
```diff
@@ -41,12 +41,14 @@
   tags:
     - smoke-tests
 
-- name: Install st2 pack from exchange
-  become: yes
+- name: Get st2 pack info
```
armab:
```suggestion
- name: Verify if st2 pack was installed
```

roles/StackStorm.st2smoketests/tasks/main.yml
punkrokk: agreed.

roles/StackStorm.st2smoketests/tasks/main.yml
punkrokk: Fine :)

Issue #249, opened 2019-08-13
Support for Ubuntu 18.04


gopivalleru:
I've ran ansible playbook for stackstorm on Ubuntu 18 which trowed multiple errors. Below are the changes I had to make to make installation work

Mongo DB changes to support 4.0:
StackStorm.mongodb/defaults/main.yml
mongodb_version: "4.0"
mongodb_apt_keys:
"3.2": "42F3E95A2C4F08279C4960ADD68FA50FEA312927"
"3.4": "0C49F3730359A14518585931BC711F9BA15703C6"
"4.0": "9da31620334bd75d9dcb49f368818c72e52529d4"

changed print command to python 3 compatible.
StackStorm.st2/tasks/version.yml
under task: Get installed st2 version
change command: /opt/stackstorm/st2/bin/python -c 'import st2common; print st2common.version'
to

command: /opt/stackstorm/st2/bin/python -c 'import st2common; print(st2common.version)'

Mistral is not supported on Ubuntu 18.04
comment or remove StackStorm.st2mistral role from stackstorm.yml role.

armab:
Thanks for sharing that info!

Could you compose the PR that will implement some of these additions for Ubuntu 18?

PR #253, opened 2019-12-16
Minor changes to support Ubuntu Bionic.

brucellino:
This makes minor changes necessary to the playbook tasks and variables
which allow them to be executed against Ubuntu Bionic (18.04) hosts.

Changes made include

1. Change MongoDB version to 4.0
2. update MongoDB repo signing keys
3. Update task "Get installed st2 version" to use python3 syntax.
4. Comment out Mistral role, since it is not supported on Ubuntu 18.04

Ref: https://github.com/StackStorm/ansible-st2/issues/249

Signed-off-by: Bruce Becker <bruce.becker@uefa.ch>

### DIFF
--- a/roles/StackStorm.st2/tasks/version.yml
+++ b/roles/StackStorm.st2/tasks/version.yml
@@ -1,7 +1,8 @@
 ---
 # Getting the current st2 version is required to understand which 'st2_services' to restart
 - name: Get installed st2 version
-  command: /opt/stackstorm/st2/bin/python -c 'import st2common; print st2common.__version__'
+  # command: /opt/stackstorm/st2/bin/python -c 'import st2common; print st2common.__version__'
+  command:  /opt/stackstorm/st2/bin/python -c 'import st2common; print (st2common.__version__)'
   changed_when: no
   check_mode: no
   register: _st2_version_installed
--- a/stackstorm.yml
+++ b/stackstorm.yml
@@ -3,12 +3,17 @@
   hosts: all
   environment: "{{ st2_proxy_env | default({}) }}"
   roles:
-    - StackStorm.mongodb
+    - role: StackStorm.mongodb
+      mongodb_version: 4.0
+      mongodb_apt_keys:
+        "3.2": "42F3E95A2C4F08279C4960ADD68FA50FEA312927"
+        "3.4": "0C49F3730359A14518585931BC711F9BA15703C6"
+        "4.0": "9da31620334bd75d9dcb49f368818c72e52529d4"
     - StackStorm.rabbitmq
     - StackStorm.postgresql
     - StackStorm.st2repo
     - StackStorm.st2
-    - StackStorm.st2mistral
+    # - StackStorm.st2mistral
     - StackStorm.nginx
     - StackStorm.st2web
     - StackStorm.nodejs


stackstorm.yml
```diff
@@ -3,12 +3,17 @@
   hosts: all
   environment: "{{ st2_proxy_env | default({}) }}"
   roles:
-    - StackStorm.mongodb
+    - role: StackStorm.mongodb
+      mongodb_version: 4.0
+      mongodb_apt_keys:
+        "3.2": "42F3E95A2C4F08279C4960ADD68FA50FEA312927"
+        "3.4": "0C49F3730359A14518585931BC711F9BA15703C6"
+        "4.0": "9da31620334bd75d9dcb49f368818c72e52529d4"
```
armab: Please move mongodb 4.0 apt key to role defaults:
https://github.com/StackStorm/ansible-st2/blob/45a463fdb4bce63be02b38624eb46d7ea7350952/roles/StackStorm.mongodb/vars/main.yml#L1-L6

stackstorm.yml
```diff
@@ -3,12 +3,17 @@
...
     - StackStorm.rabbitmq
     - StackStorm.postgresql
     - StackStorm.st2repo
     - StackStorm.st2
-    - StackStorm.st2mistral
+    # - StackStorm.st2mistral
```
armab: Good start. Now need to add condition to disable this role import for Ubuntu Bionic only.

Also keep in mind that the same condition should be applied to `postgresql` role as well as that's used by mistral only.

roles/StackStorm.st2/tasks/version.yml
```diff
@@ -1,7 +1,8 @@
 ---
 # Getting the current st2 version is required to understand which 'st2_services' to restart
 - name: Get installed st2 version
-  command: /opt/stackstorm/st2/bin/python -c 'import st2common; print st2common.__version__'
+  # command: /opt/stackstorm/st2/bin/python -c 'import st2common; print st2common.__version__'
```
armab:
```suggestion
```

Commented value could be removed.

stackstorm.yml
```diff
@@ -3,12 +3,17 @@
   hosts: all
   environment: "{{ st2_proxy_env | default({}) }}"
   roles:
-    - StackStorm.mongodb
+    - role: StackStorm.mongodb
+      mongodb_version: 4.0
```
armab: :+1: 
Similar to `StackStorm.st2mistral` import, need to add condition here so use `4.0` only for Ubuntu Bionic.

roles/StackStorm.st2/tasks/version.yml
brucellino: ah sorry, that one made it through my mistake.

stackstorm.yml
brucellino: will do.

nmaludy:
@brucellino looks like you have an ansible lint error: https://circleci.com/gh/StackStorm/ansible-st2/920?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

armab:
Implemented in https://github.com/StackStorm/ansible-st2/pull/269

PR #255, opened 2020-01-09
Improve chatops smoketest resiliency on travis

cognifloyd:
This is an attempt to make the travis tests more resilient when running the hubot tests in parallel.
~This will wait for 300 seconds for the port to close (no longer be in use) to avoid the EADDRINUSE error.~
As @armab pointed out, the EADDRINUSE error is a red herring. The real issue is slack Rate limiting. So this adds a retry until to hopefully make more CI runs pass without intervention.

### DIFF
--- a/roles/StackStorm.st2smoketests/tasks/st2chatops.yml
+++ b/roles/StackStorm.st2smoketests/tasks/st2chatops.yml
@@ -10,7 +10,7 @@
   shell: >
     set -o pipefail &&
     export n=0$(cat $attempts_count_file) &&
-    timeout $(( 9 + 5 ** n )) bash -c '(sleep $(( 4 + 5 ** n )); echo exit ) | bin/hubot';
+    timeout $(( 9 + 6 ** n )) bash -c '(sleep $(( 4 + 6 ** n )); echo exit ) | bin/hubot';
     echo $((n+1)) > $attempts_count_file
   # $a,$n = homegrown exponential backoff (increasing sleep lets the slack client lib retry automatically)
   args:
--- a/roles/StackStorm.st2smoketests/tasks/st2chatops.yml
+++ b/roles/StackStorm.st2smoketests/tasks/st2chatops.yml
@@ -1,17 +1,23 @@
 ---
+- name: Prepare temp file to track retry attempts
+  tempfile:
+    state: file
+  register: attempts_count
+  changed_when: no  # skip idempotence check
+
 - name: Verify st2chatops using bin/hubot
   # when editing, make sure it works for at least 2 adapters: 'shell' and 'slack'
   shell: >
     set -o pipefail &&
-    export a=/tmp/.hubot.{{ 10000 | random }}.attempts &&
-    export n=$(cat $a 2>/dev/null || echo 0) &&
-    timeout $(( 9 + 5 ** n )) bash -c '(sleep $(( 4 + 5 ** n )); echo exit ) | bin/hubot' &&
-    echo $((n+1)) > $a
+    export n=0$(cat $attempts_count_file) &&
+    timeout $(( 9 + 5 ** n )) bash -c '(sleep $(( 4 + 5 ** n )); echo exit ) | bin/hubot';
+    echo $((n+1)) > $attempts_count_file
   # $a,$n = homegrown exponential backoff (increasing sleep lets the slack client lib retry automatically)
   args:
     chdir: /opt/stackstorm/chatops/
     executable: /bin/bash
   environment:
+    attempts_count_file: "{{ attempts_count.path }}"
     HUBOT_LOG_LEVEL: debug
   register: hubot_output
   failed_when: no
@@ -24,6 +30,12 @@
   retries: 3
   # random delay so that retries accross parallel runs don't perfectly align
   delay: "{{ 15 + (7 | random) }}"
+
+- name: Cleanup temp file to track retry attempts
+  file:
+    path: "{{ attempts_count.path }}"
+    state: absent
+  changed_when: no  # skip idempotence check
 
 # Additional task to provide better error message
 - name: Fail if st2chatops couldn't load st2 commands
--- a/roles/StackStorm.st2smoketests/tasks/st2chatops.yml
+++ b/roles/StackStorm.st2smoketests/tasks/st2chatops.yml
@@ -1,7 +1,7 @@
 ---
 - name: Verify st2chatops using bin/hubot
   # when editing, make sure it works for at least 2 adapters: 'shell' and 'slack'
-  shell: set -o pipefail && timeout 10 bash -c '(sleep 5; echo exit ) | bin/hubot'
+  shell: "set -o pipefail && timeout {{ hubot_timeout }} bash -c '(sleep {{ hubot_sleep }}; echo exit ) | bin/hubot'"
   args:
     chdir: /opt/stackstorm/chatops/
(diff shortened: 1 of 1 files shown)


cognifloyd:
@armab I think this should reduce the impact of slack rate limiting. wdyt?

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
```diff
@@ -11,6 +11,14 @@
   failed_when: no
   changed_when: no
```
armab: Blank line could be removed here
```suggestion
```

armab:
Yes, this looks like a good use of retry with delays. Should definitely improve it :+1:

cognifloyd:
It looks like this actually made it worse.

cognifloyd:
K. I think this is ready now. I updated it to have a more exponential-ish back off, and randomized the delay between retries slightly.

cognifloyd:
Hmm. I finally tripped the retry logic https://travis-ci.org/StackStorm/ansible-st2/builds/635053348
1 build succeeded first try
7 builds triggered the retry logic
3/7 succeeded
4/7 failed

So it's not quite right yet. The command is only templated once, so using attempts to simulate an exponential backoff doesn't work.

cognifloyd:
With the version in 661b037:
https://travis-ci.org/StackStorm/ansible-st2/builds/635132037
1 build succeeded on the first try
7 triggered the retry logic
5/7 succeeded
2/7 failed
https://travis-ci.org/StackStorm/ansible-st2/builds/635132304
8 builds triggered the retry logic
7/8 succeeded
1/8 failed

So, this is much more reliable. The latest commit b0f9695 extends the sleep (only during retries) to hopefully catch the last few builds hitting the rate limit. This is already much better, and with this extra bit of sleep, I think it's good enough. Hopefully we won't have to kick travis as much.

Note that if a build succeeds on the first try, then it should be just as it is now (timeout 10 and sleep 5). Those times are only increased for retries.

cognifloyd:
That is very disappointing. I seem to trigger these errors every time I try to rebase my PRs, and then they sit around with failed CI again.

The travis auto-cancel feature has been problematic for me on other projects. It will only cancel builds that have NOT started building. So if nothing is running, you update a PR, it starts building, and then you update the PR again a few minutes later, that first build will still have to complete before the next one starts. So, auto-cancel might reduce some pressure, but I doubt it'll make much difference.

cognifloyd:
Note that the Build History (triggered on a schedule or by merges to master) will only have failures if PRs get pushed at the same time as one of those builds happens to run. Because they are infrequent enough, you will rarely see errors there. The parallel build pressure is only a significant issue for PR authors. To see that history you need to look under "Pull Requests".
Almost all of the failures in this screenshot are due to slack rate limiting errors, not actual errors with the PR:

image

cognifloyd:
It looks like you've lowered the max concurrent jobs from 4 to 2. :+1: Thanks! Hopefully jobs won't error on me so often.

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
```diff
@@ -1,15 +1,41 @@
...
+  changed_when: no  # skip idempotence check
+
 - name: Verify st2chatops using bin/hubot
   # when editing, make sure it works for at least 2 adapters: 'shell' and 'slack'
-  shell: set -o pipefail && timeout 10 bash -c '(sleep 5; echo exit ) | bin/hubot'
+  shell: >
+    set -o pipefail &&
+    export n=0$(cat $attempts_count_file) &&
+    timeout $(( 9 + 6 ** n )) bash -c '(sleep $(( 4 + 6 ** n )); echo exit ) | bin/hubot';
+    echo $((n+1)) > $attempts_count_file
```
nmaludy: would love to see some comments here on how this works, reading this it's pretty confusing and not very intuitive as to what it's doing.

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
```diff
@@ -1,15 +1,41 @@
 ---
+- name: Prepare temp file to track retry attempts
```
nmaludy: maybe at the top of this file, some comments in a summary of how this retry functionality works would be useful (what the temporary file is for, etc)

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
```diff
@@ -1,15 +1,41 @@
...
+  tempfile:
+    state: file
+  register: attempts_count
```
nmaludy: does this initialize attempts_count to 0 on the first run?

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
```diff
@@ -1,15 +1,41 @@
...
   failed_when: no
   changed_when: no
+  # any EADDRINUSE error in stdout is normal (an artifact of the js express http framework)
+  # On Travis, with failures when running many builds in parallel, the real error message is:
+  # "Rate limited, will retry # seconds". This comes from @slack/client (used by hubot-slack).
+  # On success (with Slack), the log says: 'SlackBot#authenticated() Found self in RTM start data'
+  until: "'Rate limited, will retry' not in hubot_output.stdout_lines | last"
+  retries: 3
+  # random delay so that retries accross parallel runs don't perfectly align
+  delay: "{{ 15 + (7 | random) }}"
```
nmaludy: probably a expression for randomness would be: `{{ range(0, 30) | random }}` or something along those lines (pick any min/max you like) 

if i look at the way that `random()` is implemented it generates a random number and picks one from the sequence (list) that is passed in. Since you're always passing in 7, it is always going to pick the number 7 and add 15.

nmaludy:
@cognifloyd did you mean to commit the file `asdf` ?

cognifloyd:
> @cognifloyd did you mean to commit the file `asdf` ?

LOL. no.

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
cognifloyd: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#random-number-filter
> To get a random number between 0 and a specified number:
> ```
> "{{ 60 | random }} * * * * root /script/from/cron"
> # => '21 * * * * root /script/from/cron'
> ```

It's an ansible-specific extension on top of the default jinja filter.
https://github.com/ansible/ansible/blob/v2.9.4/lib/ansible/plugins/filter/core.py#L226-L231

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
cognifloyd: I should rename that ansible var to attempts_count_file to mirror the var name I'm using in Bash and make the var easier to follow.
If the file is empty, it defaults to 0 because of the 0 before the cat here:
`n=0$(cat $attempts_count_file)`

roles/StackStorm.st2smoketests/tasks/st2chatops.yml
cognifloyd: It was confusing to write too. Yeah, I'll add some comments.

cognifloyd:
I just rebased, removed the `asdf` file (oops), and added comments to explain how the backoff algorithm works.

PR #256, opened 2020-01-29
Configure/Set username/password for Mongo

cognifloyd:
Adds auth to mongo when the `mongodb_auth_enable` flag is true.
Any additional users should be passed in via `mongodb_users`.
There are several other default variables as well (like `mongodb_host`,
`mongodb_port`) that can be overridden in the play or inventory that uses
this role.

This only adds users if explicitly requested. External tasks/roles can
also import the mongodb_auth.yml tasks to add users as required after
mongo is installed.

This uses lineinfile to edit mongod.conf with regexs to catch as many
edge cases in yaml formatting (spaces, quotes) as possible.
Also, this uses a bit of python to validate that the yaml file was
modified in such a way that it is still valid yaml, and the entries
introduced in the file are present as expected.

This uses the mongo shell to see if authorization is required before
adding any users. Before adding users, including admin, we need to know
if authentication is already enabled in the running instance of mongodb
to (a) have an idempotent playbook, and (b) to cleanly handle edge cases
where people are configuring a mongo instance that is already has auth
configured or partially configured.

For a truly idempotent playbook, the playbook needs to be able to run
both before authentication is enabled and after it is enabled. The check
validates the state of mongo auth including: Is auth enabled? Are users
configured?. Even after mongo is restarted with auth enabled, the check
task will still return rc=0 until users are added due to the localhost
exception[1].

As explained in code comments, we only update mongo user passwords
on_create (vs always) because of a mongodb restriction that prevents the
ansible module from idempotently/sanely setting the password. The role
allows overriding that by setting mongodb_force_update_password.

Part of #75 (Configure/Set username/password for Mongo)

[1] https://docs.mongodb.com/manual/core/security-users/#localhost-exception

### DIFF
--- a/roles/StackStorm.mongodb/defaults/main.yml
+++ b/roles/StackStorm.mongodb/defaults/main.yml
@@ -2,3 +2,27 @@
 # MongoDB version to install
 # Should be '3.2' or '3.4'
 mongodb_version: "3.4"
+
+# do not enable mongo auth by default. Make sure to add required users in mongodb_users if auth is enabled.
+mongodb_enable_auth: no
+
+# when adding auth, the login credentials to use
+mongodb_host: 127.0.0.1
+mongodb_port: 27017
+mongodb_admin_username: admin
+# For production use - please change the admin password!
+mongodb_admin_password: "{{ mongodb_default_admin_password }}"
+# The default is separate so the st2 role can provide a default without overriding a user provided password.
+mongodb_default_admin_password: "Ch@ngeMe"
+
+# Additional users to add.
+mongodb_users: []
+#  - db: st2
+#    username: st2mongo
+#    password: "Ch@ngeMe"
+#    roles: readWrite
+
+# whether or not to force a password update for any users in mongodb_users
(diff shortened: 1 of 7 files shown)


roles/StackStorm.mongodb/defaults/main.yml
```diff
@@ -2,3 +2,27 @@
 # MongoDB version to install
 # Should be '3.2' or '3.4'
 mongodb_version: "3.4"
```
punkrokk: I think that we should install 3.4 if EL7 or EL6.

cognifloyd: This is installing 3.4 for all platforms that ansible-st2 supports right now. The only way it would be something else is if someone set the mongodb_version var. But with bionic / EL8 etc we'll need some more intelligent version selection.

That said, selecting other versions of mongo would be a different PR.

cognifloyd:
This PR is an orphan if anyone wants to pick it up, go for it.

PR #258, opened 2020-04-13
Custom nginx

punkrokk:
Adds `st2web_nginx_conf` to expose a configuration option to provide a custom `nginx.conf` file. 

Defaults to: `/usr/share/doc/st2/conf/nginx/st2.conf`

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | **st2web**
 | `st2web_ssl_certificate`     | `null` | String with custom SSL certificate (`.crt`). If not provided, self-signed certificate will be generated.
 | `st2web_ssl_certificate_key` | `null` | String with custom SSL certificate secret key (`.key`). If not provided, self-signed certificate will be generated.
-| `st2web_nginx_conf`     | `/usr/share/doc/st2/conf/nginx/st2.conf` | String with path to nginx conf file.
+| `st2web_nginx_config`     | `null` | String with a custom nginx configuration file (`st2.conf`). If not provided, the default st2.conf will be used.
 | **ewc**
 | `ewc_license`            | `null`        | EWC license key is required for installing EWC enteprise bits via this ansible role.
 | `ewc_repo`               | `enterprise`  | EWC PackageCloud repository to install. [`enterprise`](https://packagecloud.io/StackStorm/enterprise/), [`enterprise-unstable`](https://packagecloud.io/StackStorm/enterprise-unstable/), [`staging-enterprise`](https://packagecloud.io/StackStorm/staging-enteprise/), [`staging-enterprise-unstable`](https://packagecloud.io/StackStorm/staging-enterprise-unstable/)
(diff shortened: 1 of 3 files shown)


roles/StackStorm.st2web/tasks/main.yml
```diff
@@ -66,8 +66,8 @@
   become: yes
   file:
     state: link
-    src: /etc/nginx/sites-available/st2.conf
-    dest: /etc/nginx/sites-enabled/st2.conf
+    src: /etc/nginx/sites-available/"{{ st2web_nginx_conf | basename }}"
```
nmaludy, line 69: should it always be called st2.conf?

this way the custom one will overwrite the existing if it exists?

punkrokk: hm. Maybe? I like having them different personally - but I can see your utility...

README.md
```diff
@@ -53,6 +53,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | **st2web**
 | `st2web_ssl_certificate`     | `null` | String with custom SSL certificate (`.crt`). If not provided, self-signed certificate will be generated.
 | `st2web_ssl_certificate_key` | `null` | String with custom SSL certificate secret key (`.key`). If not provided, self-signed certificate will be generated.
+| `st2web_nginx_conf`     | `/usr/share/doc/st2/conf/nginx/st2.conf` | String with path to nginx conf file.
```
armab, line 56: I'm suspecting that exposing `st2.conf` nginx config this way will lead to confusion or wrong/inappropriate use.

Also taking the var description `String with path to nginx conf file.`, as as user I'd also wonder why this setting is set to `/usr/share/doc/..` and why it's path to nginx config file or how nginx will read config from that dir (from the var description).

---

In a normal situation I'd say transferring custom st2.conf nginx config to some specific place like `/tmp/st2.conf` and overriding this setting in `ansible-st2` looks a bit strange as you'll need another Ansible task to upload that file to VM prior to running ansible-st2.

Instead of conf file path on a VM, how about exposing this var as nginx raw config/text?
This would be somewhat similar to `st2web_ssl_certificate_key` and `st2web_ssl_certificate` approach.

punkrokk: @arma So are you saying, 

1. leave it the way it was if null, and write it if not null?
Or 
2. Ship the standard file in ansible, and if that var is set, then copy the contents?

punkrokk: I found an approach.

roles/StackStorm.st2web/defaults/main.yml
```diff
@@ -7,3 +7,6 @@ st2web_ssl_certificate: null
 
 # String with custom SSL certificate private key. If not provided, self-signed certificate will be generated.
 st2web_ssl_certificate_key: null
+
+# A string that represents an custom nginx config file
```
armab, line 11: The var description in README.md is pretty good :+1: 
Please include that var description here as well.

roles/StackStorm.st2web/tasks/main.yml
```diff
@@ -61,6 +61,17 @@
   notify:
     - restart nginx
   tags: st2web
+  when: not st2web_nginx_config
+
+- name: Copy custom Nginx config
```
armab, line 66: Would be good then to modify previous section to: `name: Copy default Nginx config`

armab:
This looks damn good and consistent now :+1: 
Please also check the ansible-lint warnings.

PR #261, opened 2020-06-12
Ansible-playbooks to install on Centos8

### LINKED ISSUE #248: Add support for CentOS / RHEL 8
To go along with StackStorm/st2#4761

Once this is done and there are CentOS 8 packages available, we can work to get this implemented in Ansible.

Things to handle:

- [ ] Mongodb 4.0
- [ ] Node 10
- [ ] CentOS 8 vs RHEL 8
- [ ] Removed some unused packs which were installed in EL7
- [ ] Mistral and Postgres no longer installed
- [ ] Python3 install numbered version

### Updates for EL8:
- [ ]  RabbitMQ on RHEL 8 requires: `sudo yum -y module enable perl:5.26`
- [ ]  Currently need the rabbit packagecloud repo:
```
# Install rabbit from packagecloud
  curl -s https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh | sudo bash
  sudo yum makecache -y --disablerepo='*' --enablerepo='rabbitmq_rabbitmq-server'
```
- [ ] selinux policy utils package name is different: `policycoreutils-python -> policycoreutils-python-utils`
- [ ]  modify `[includes/common.sh](https://github.com/StackStorm/st2-packages/blob/f0e1e650d17ccee9c8e4ab05bfdbd1a664378089/scripts/st2bootstrap-el8.sh#L256-L260)` to force `PEM` key type. 
- [ ]  MongoDB to 4.0 (Readme not updated yet)

See https://github.com/StackStorm/st2docs/pull/955 more differences from EL7 -> EL8.

amanda11:
Closes #248

I have tested this on Centos8/7/Xenial with EWC licenses.

Some comments / queries on implementation:

1.  Many roles do an import based on package manager, therefore expect to find a _dnf.yml file. I have included those files. The alternative would be to change the imports so that if the package manager was dnf it still imported the yum version.
Do we want to have dnf and yum files? 
Where the RH8 versus 6/7 differ then it omits the need for checks on RH version in the _yum variant, but it also means in some cases the dnf and yum files are identical.

2. I didn't manage to test the firewall-cmd as on my Centos8 not enabled. But have used same logic as single node install. 

3. I created a mongodb4 variable. You can't have conditional defaults so if just have the one default mongodb_version then for installing on RH8 then it would default to 3.4. 
An alternative might have been to have some conditional logic that if the mongodb version is < 4 then to set to the default mongodb4 version instead?

4. In the mongodb dependencies then for RH8 I have removed the dependency on python-ndg_httpsclient. It isn't available on the RH8 manual install. All other python imports are changed to their python3 one.


stackstorm.yml
```diff
@@ -5,10 +5,12 @@
...
     - StackStorm.mongodb
     - StackStorm.rabbitmq
-    - StackStorm.postgresql
+    - role: StackStorm.postgresql
+      when: not (ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '8')
     - StackStorm.st2repo
     - StackStorm.st2
-    - StackStorm.st2mistral
+    - role: StackStorm.st2mistral
+      when: not (ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '8')
```
armab, line 13: :+1:

roles/StackStorm.st2repo/tasks/st2repo_dnf.yml
```diff
@@ -0,0 +1,41 @@
...
+    section: main
+    option: tsflags
+    value: nodocs
+    state: absent
+  when: ansible_facts.os_family == "RedHat"
+  tags: st2repo
+
+- name: Add StackStorm repo
+  become: yes
+  yum_repository:
```
armab, line 29: Yeah, indeed many `_dnf.yml` files are identical to `_yum.yml` files.
We definitely want to avoid code duplication and use one set of rules to manage both.

roles/StackStorm.st2chatops/tasks/main.yml
```diff
@@ -1,4 +1,18 @@
...
+  register: http_parser
+  tags: st2chatops
+
+- name: Install http-parser if needed
+  yum:
+      name: http://repo.okay.com.mx/centos/8/x86_64/release//http-parser-2.8.0-2.el8.x86_64.rpm
+      state: present
+  when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '8' and (http_parser.results|length == 0)
+  tags: st2chatops
+
```
armab, line 15:
```suggestion
```
This could be removed.

Per https://bugs.centos.org/view.php?id=13669 the bug is not related to `el8`, @punkrokk copy-pasted it by mistake when added el8 support.
I'm still waiting him to fix that in https://github.com/StackStorm/st2-packages/blob/29a493e33de928e6df452a9102e80aaeadc5ec53/scripts/st2bootstrap-el8.sh#L694-L697 :wink:

README.md
```diff
@@ -13,6 +13,7 @@ Aka IFTTT orchestration for Ops.
 * Ubuntu Xenial (16.04)
 * RHEL6 / CentOS6
 * RHEL7 / CentOS7
+* RHEL8 / CentOS8
```
armab, line 16: Please try to also add CentOS8 builds in `.kitchen.yml` and `.travis.yml` so we'll get new OS support tested in this PR.

README.md
```diff
@@ -102,6 +103,7 @@ These are the platforms we must support (must pass end-to-end testing):
 - Trusty
 - CentOS6
 - CentOS7
+- CentOS8
```
armab, line 106: Take a look at Ansible meta files https://github.com/StackStorm/ansible-st2/blob/master/roles/StackStorm.st2/meta/main.yml#L13-L16 where new `EL8` could be added too.

roles/StackStorm.st2repo/tasks/st2repo_dnf.yml
punkrokk: I agree with @armab's suggestion

roles/StackStorm.nginx/tasks/nginx_yum.yml
```diff
@@ -66,3 +124,31 @@
...
+     path: /sys/hypervisor/uuid
+     line: "ec2"
+     state: present
+  check_mode: yes
+  ignore_errors: true
+  register: ec2
+  tags: nginx
+  when: (ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '8')
+
+- name: Add http and https services to firewalld
```
punkrokk, line 140: I think we want to ensure firewalld is not present in ec2, right?

amanda11: In the one line install it only enabled httpd and https services if it was RH8 and there was a hypervisor/sysuuid and that did not contain ec2
I did wonder about just changing to be enable those services if firewalld was enabled, as they seemed more what was needed.

amanda11:
I need to fix 5 'changed' tasks on the idempotent. Trying to figure out if there is a yum module <disable> where you can specify the stream, so that it doesn't get altered. Just doing some research on this to fix the travis idempotency errors reported.

roles/StackStorm.nginx/tasks/nginx_yum.yml
```diff
@@ -40,10 +44,64 @@
...
+- name: Cleanup double comments
+  become: yes
+  replace:
+     regexp: "##"
+     replace: "#"
+     path: /etc/nginx/nginx.conf
+  tags: nginx
+  when: (ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '8')
+
+- name: Cleanup comment closing out server block
```
armab, line 92: These set of tasks coming from bash needs a proper re-write by using Ansible powers better.
One-line bash/awk/tmp hacks are coming from the fact it was the best way to do it for curl-bash installer.

The objective here is simple: disable default nginx record which is conflicting with the st2 nginx config.

roles/StackStorm.mongodb/tasks/mongodb_dnf.yml
```diff
@@ -0,0 +1,54 @@
+---
+- name: yum | Install mongodb dependencies
```
armab, line 2: There looks like still `mongodb_dnf.yml` file still present. Probably missed to remove in previous iteration.

amanda11: I left that one in as every single task was different between RH7 and RH8 for mongo. I could keep as a single _yum.yml but then each task will need to import RH7/RH8 specific variables to know what to import. But can go that way if preferred.

roles/StackStorm.nginx/tasks/nginx_yum.yml
amanda11: > These set of tasks coming from bash needs a proper re-write by using Ansible power better.
> One-line bash/awk/tmp hacks are coming from the fact it was the best way to do it for curl-bash installer.
> 
> The objective here is simple: disable default nginx record which is conflicting with the st2 nginx config.

Have that fixed replaced with 2 replace calls instead (couldn't get a single one to do it). Need to fix some idempotent issues with the yum module enable/disable calls which aren't supported in ansible yum's module. Will put up changes once I've got those fixed and re-run. Probably early next week now though.

punkrokk: @amanda11 - this is kick ass! thanks for your help on this!

amanda11:
I still have 2 idempotent fixes to do on the yum module enable/disable. But wanted to push to get the travis jobs to run to confirm that its the only problem after the last refactor.

Other thoughts:
1. mongodb_dnf.yml - there isn't any common tasks with mongodb_yum.yml - but I can probably refactor it to one mongodb_yum.yml and different variable includes instead. As the tasks are the same just differnet variables. Will try and do refactor of that next week unless prefer to keep as is.
2. On the firewall-cmd the single line install just seems to enable http/https commands if firewalld is enabled and there is a hypervisor/uuid file and it does NOT contain ec2. I've just used the same condition but I don't know if it could be simplifieid to just check if firewalld is enabled? (Not sure on the history of that one...) (on my Centos8 firewalld is NOT enabled)

roles/StackStorm.st2/tasks/auth.yml
```diff
@@ -11,12 +11,16 @@
   until: _task is succeeded
   when: ansible_facts.os_family == 'Debian'
 
+- name: Include RedHat OS-specific variables
+  include_vars: st2_{{ ansible_facts.pkg_mgr }}.yml
```
armab, line 15: Following the logic, EL8 uses `python3-passlib`, while older distros (EL7 and EL6) `python-passlib`.

I think more correct construct would be the following:

```yaml
- name: Include OS specific variables
  include_vars: "{{ item }}"
  with_first_found:
    - "{{ ansible_os_family|lower }}_{{ ansible_distribution_major_version }}.yml"
    - "{{ ansible_os_family|lower }}.yml"
```

`redhat_8.yml`
`redhat.yml`

As it's not really `dnf` vs `yum` thing, but OS version and platform specific logic.

roles/StackStorm.mongodb/tasks/mongodb_dnf.yml
armab: Yes, having el7/el8 specific variables to import is a widely used practice and would look more clear. Just included an  example in other comment.

amanda11:
@armab thanks for feedback, will
Implement those comments - will make it neater.

amanda11:
@armab @punkrokk Have implemented your comments, let me know any other feedback. All checks now passing.

amanda11:
Thanks. I believe if I disable / enable on the yum install itself then the problem is any later yum update would then take from AppStream and not nodejs. So you do have to do disable the stream separately for the module.

armab:
Yes, right.
I've checked the upgrade case and the fact that `st2chatops` package relies on `epoch:2` of nodejs (https://github.com/StackStorm/st2chatops/blob/master/rpm/st2chatops.spec#L11), `nodesource` version will be always used as a dependency.

* Appstream provides: `1:10.19.0-2.module_el8.2.0+310+695ca693`
* Nodesource provides: `2:10.21.0-1nodesource`

So while we're on the safe side, I'm good to keep it as an additional safety check because I found similar logic in the official nodesource script `curl -sL https://rpm.nodesource.com/setup_10.x | sudo -E bash -` script:
```
## As yum will try to install Node.js from the AppStream repository
instead of the NodeSource repository, the AppStream's version of Node.js has to be disabled.
## Run `sudo yum module enable -y nodejs` to reactivate the AppStream's Node.js repository.

+ yum module disable -y nodejs
```

roles/StackStorm.nginx/tasks/nginx_redhat.yml
```diff
@@ -0,0 +1,123 @@
...
+- name: Adjust SELinux to allow network access for nginx
+  become: yes
+  seboolean:
+    name: httpd_can_network_connect
+    state: yes
+    persistent: yes
+  when: ansible_facts.selinux.status == "enabled" and ansible_facts.selinux.mode == "enforcing"
+  tags: nginx
+
+- name: Discover if hypervisor is EC2
```
armab, line 97: OK, based on @amanda11 comment that we have some AWS EC2-specific rules to enable/disable firewall rules looks strange to me.

The original code based on @punkrokk initial work from the scripted installer:
https://github.com/StackStorm/st2-packages/blob/29a493e33de928e6df452a9102e80aaeadc5ec53/scripts/st2bootstrap-el8.sh#L685-L690

@punkrokk Can you shed more light about this case and why we don't want to change the firewall rules for AWS? How about other VMs, other cloud providers, bare-metal servers, etc? What's so specific about it?

meta/main.yml
```diff
@@ -22,6 +22,7 @@ galaxy_info:
       versions:
         - 6
         - 7
+        - 8
```
armab, line 25: Thanks!
Can you please also add similar meta information for other affected roles?

Check `roles/StackStorm.*/meta/main.yml` files.

roles/StackStorm.nginx/tasks/nginx_redhat.yml
punkrokk: @armab @amanda11 Standard practice on AWS is to not run a firewall on the EC2 instance and rely on security groups. They are functionally the same thing, while security groups are much more flexible and centrally manageable. 

Most AMIs in EC2 ship without a host based firewall enabled. The reason this ended up in the Ansible playbook here is because I was testing on the official RHEL AMI, and this was a failure point in EC2. 

I think we may want to have other checks for GCP and Azure, but that should probably be a second ticket.

armab: OK, so was the original failure due to absense of the `firewall-cmd` binary or what?

punkrokk: Yes it was caused by the absence of the firewall-cmd binary. And I don't believe we should install it.

armab: OK.
So sounds like original scripted installer instead of checking for `EC2` hypervisor, should check for `firewall-cmd` binary presence instead.

armab: I also agree we shouldn't try to install the binary.

@punkrokk Based on this discussion, can you please modify the original https://github.com/StackStorm/st2-packages/blob/29a493e33de928e6df452a9102e80aaeadc5ec53/scripts/st2bootstrap-el8.sh#L685-L690 scripted installer and check for `firewall-cmd` binary presence instead of ec2 to avoid such errors and confusion in future?

This will also make the rule work outside of EC2 in any provider which makes more sense.

punkrokk: Maybe. But first, consider a situation where we desire to have firewall-cmd installed, outside of a cloud instance, and it isn't. How would we handle that?

punkrokk:
I would recommend we now test this on CentOS 8.2, since it got released yesterday and this work is fresh in everyone's mind.

roles/StackStorm.nginx/tasks/nginx_redhat.yml
armab: If there is a `firewall-cmd` present, - we always want to allow http and https to make st2web effectively work, doesn't matter cloud/VM/bare metal/etc.
If there is no `firewall-cmd` binary, - don't try to do anything.

What's the other case? I don't understand the statement about desire to have firewall-cmd installed. Can you clarify in more detail the scenario?

punkrokk: So - I think this might open a can of worms. `firewalld` (like `iptables`) both are cli interfaces to the Linux Netfilter packet manipulation system. The best explanation of this I can find at the moment is: https://medium.com/@iced_burn/compare-firewalld-iptables-nftables-netfilter-de08a8d21b5b.

So - where I am stuck, is that just because `firewalld` or `iptables` is or isn't installed, that (in my mind) doesn't guarantee state. So, we could have a box that has iptables installed and a `deny all` on port 80 and 443, but everything will seem fine. 

At the end of the day, I think we all are in agreement that these ansible modules are desired to be applied to a standard (as in official ISO/AMI/etc) installation of a particular OS.

That said, we will and do have situations (testing this repo on AWS for example) that will necessitate certain things. 

I believe we can take the firewalld approach, but we should note that it's not as idempotent as we may believe, since the firewall can be modified with other tools (iptables/ufw/etc).

nmaludy: FYI the approach we take in the Puppet module is that we do NOT manage the firewall. This is true for almost all puppet modules (few exceptions here and there).

The end-user can integrate then utilize our module to install the StackStorm bits that is responsible for, but leaves the rest of the system integration up to the end-user. 

The reasoning for this is similar to what @punkrokk said above. There are just too many variations with regards to firewall configurations, especially cross platform. From a Puppet perspective the only way you can guarantee configuration consistency is to have Puppet manage all rules in the firewall and purge (remove) any non-Puppet managed rules. Few organizations actually do this.

I would think of this similar to things like monitoring, authentication, centralized logging, etc. We don't and probably shouldn't worry about configuring these bits inside of our modules. Leave that to the end-user where those configurations are going to be unique to each end-user.

The only option/suggestion would be is to provide an "opt-in" flag where you can specify `manage_firewall` that would enable/disable firewall rule management and then a `firewall_type` or something along those lines. Depending on the `firewall_type` we could run the `firewall-cmd` parts. This way the user can choose the right option for them, and if they dont' like the way we configure the firewall, they can configure the firewall themselves.

armab: :+1: 
If user wants to manage the firewall rules, - they should use another Ansible role dedicated for that. This will also keep our roles cleaner.

punkrokk: ok - so what approach did you agree to @armab ? :P

punkrokk: > I also agree we shouldn't try to install the binary.
> 
> @punkrokk Based on this discussion, can you please modify the original https://github.com/StackStorm/st2-packages/blob/29a493e33de928e6df452a9102e80aaeadc5ec53/scripts/st2bootstrap-el8.sh#L685-L690 scripted installer and check for `firewall-cmd` binary presence instead of ec2 to avoid such errors and confusion in future?
> 
> This will also make the rule work outside of EC2 in any provider which makes more sense.

How do the more recent comments apply to your ask @armab

armab: @nmaludy had a good point. The same approach is followed by Chef. Ansible is no diff here with the separation of concerns via roles.

@punkrokk No firewall rule should be managed in nginx role so we'll just remove the `EC2 check` and `firewall-cmd` tasks alltogether. If user needs to manage firewall, - he should use dedicated role to do that, depending on his environment and toolset.

We also didn't add any firewall-management tasks before in nginx role.

punkrokk: Ok. That works. I’d like to see the Readme update with a warning and a breaking change added to the change log.  I appreciate this direction and agree with it, but There will be some that miss this and will come for support.

armab: Can you clarify about the Readme update?
We didn't have firewall rules in Ansible roles before, where is the breaking change?

amanda11: So I will try and address the comments tomorrow, but it’s agreed that I shall just remove those firewall-cmd calls in the ansible playbook. 
I will address the other comment. I will take a look and see if I can find somewhere to update in docs. I presume we want to point out that on distributes where firewalld is enabled they will need to ensure that http and https are enabled.

roles/StackStorm.st2chatops/tasks/main.yml
armab: Here is the follow-up fix for that: https://github.com/StackStorm/st2-packages/pull/651
cc @punkrokk

roles/StackStorm.nginx/tasks/nginx_redhat.yml
armab: Cool :+1: 
And as a follow-up for scripted curl-bash installer: https://github.com/StackStorm/st2-packages/pull/650 so we have this logic improved based on discussion.
cc @punkrokk @nmaludy

punkrokk: Maybe not a breaking change - but a warning about making sure you firewall is allowing traffic. You are right it’s not breaking

README.md
```diff
@@ -13,9 +13,12 @@ Aka IFTTT orchestration for Ops.
 * Ubuntu Xenial (16.04)
 * RHEL6 / CentOS6
 * RHEL7 / CentOS7
+* RHEL8 / CentOS8
 
 > If you're using the provided Vagrantfile, note that it uses Xenial by default.
 
+> If you're installing on RHEL8 / CentOS8 and firewalld is enabled, then you must ensure that the http and https services are enabled. The ansible-st2 roles do not configure the firewalld rules.
```
armab, line 20: An improvement to this would be having a reminder to users mentioning firewall rules they need to manage in order to open st2web access. The reason is that EL8 and very specific firewall tool is not the case here. Same could be applied to any OS with whatever firewall system it runs. 

For example: 
> In order to access StackStorm Web UI, please don't forget to ensure that `http`/`https` ports are opened in your firewall system.

And so this could be applied to any OS, any firewall system is it internal or cloud-layer firewall.

roles/StackStorm.nodejs/tasks/nodejs_redhat.yml
```diff
@@ -19,6 +19,28 @@
   register: nodesource_repo_rm
   tags: nodejs
 
+- name: Determine if nodejs disabled on AppStream for RH8
```
punkrokk, line 22: Super picky - but let use EL8?

```suggestion
- name: Determine if nodejs disabled on AppStream for EL8
```

punkrokk:
Great job @amanda11

armab:
The new EL8 support feature should be now available as [`v1.5.0`](https://github.com/StackStorm/ansible-st2/releases/tag/v1.5.0) in ansible galaxy:
https://galaxy.ansible.com/StackStorm/stackstorm

Issue #262, opened 2020-06-24
Installation failed on RH 7 - st2chatops service not found


dove-young:
Here is my installation command 

```
ANSIBLE_FORCE_COLOR=true ansible-playbook stackstorm.yml -i ./hosts  -vvv 2>&1 | tee install-2.log
```

Ansible-ST2 HEAD is 

```
commit 97108985f3351d696d3413f2db3cf3d43eb43a89
Merge: 86a1cec 35abf1d
Author: JP Bourget <jp.bourget@gmail.com>
Date:   Tue Apr 14 13:40:42 2020 -0400

    Merge pull request #258 from StackStorm/custom_nginx

    Custom nginx

```

And installation failed at here

```
TASK [StackStorm.st2chatops : Check if any adapter is enabled] *******************************************************************************************************************************************
task path: /root/ansible-st2/roles/StackStorm.st2chatops/tasks/main.yml:123
  fatal: [9.46.74.243]: FAILED! => {
    "changed": false,
    "cmd": [
        "grep",
        "-e",
        "^export HUBOT_ADAPTER=",
        "/opt/stackstorm/chatops/st2chatops.env"
    ],
    "delta": "0:00:00.008895",
    "end": "2020-06-23 19:03:40.050882",
    "invocation": {
        "module_args": {
            "_raw_params": "grep -e '^export HUBOT_ADAPTER=' /opt/stackstorm/chatops/st2chatops.env",
            "_uses_shell": false,
            "argv": null,
            "chdir": null,
            "creates": null,
            "executable": null,
            "removes": null,
            "stdin": null,
            "stdin_add_newline": true,
            "strip_empty_ends": true,
            "warn": true
        }
    },
    "msg": "non-zero return code",
    "rc": 1,
    "start": "2020-06-23 19:03:40.041987",
    "stderr": "",
    "stderr_lines": [],
    "stdout": "",
    "stdout_lines": []
}
...ignoring

TASK [StackStorm.st2chatops : Set variable enabled_adapter] **********************************************************************************************************************************************
task path: /root/ansible-st2/roles/StackStorm.st2chatops/tasks/main.yml:130
fatal: [9.46.74.243]: FAILED! => {
    "msg": "The task includes an option with an undefined variable. The error was: list object has no element 1\n\nThe error appears to be in '/root/ansible-st2/roles/StackStorm.st2chatops/tasks/main.ym
[...]

PR #263, opened 2020-06-24
fix issue: Installation failed on RH 7 - st2chatops service not found

dove-young:
Fix issue https://github.com/StackStorm/ansible-st2/issues/262

```
- name: daemon-reload
  become: yes
  command: systemctl daemon-reload
  
```

### DIFF
--- a/roles/StackStorm.st2chatops/tasks/main.yml
+++ b/roles/StackStorm.st2chatops/tasks/main.yml
@@ -202,6 +202,10 @@
   become: yes
   command: systemctl daemon-reload
   
+- name: enable st2chatops
+  become: yes
+  command: systemctl enable st2chatops
+  
 - name: Ensure st2chatops service is enabled and running
   become: yes
   service:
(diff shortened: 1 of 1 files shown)


roles/StackStorm.st2chatops/tasks/main.yml
```diff
@@ -198,6 +198,14 @@
   notify:
     - restart st2chatops
 
+- name: daemon-reload
+  become: yes
+  command: systemctl daemon-reload
+  
+- name: enable st2chatops
+  become: yes
+  command: systemctl enable st2chatops
```
armab, line 207: The service is already enabled by the task below via https://docs.ansible.com/ansible/latest/modules/service_module.html

roles/StackStorm.st2chatops/tasks/main.yml
```diff
@@ -198,6 +198,14 @@
   notify:
     - restart st2chatops
 
+- name: daemon-reload
+  become: yes
+  command: systemctl daemon-reload
```
armab, line 203: You can try to use Ansible's module https://docs.ansible.com/ansible/latest/modules/systemd_module.html for daemon-reload.

Simple command is discouraged. This task will also break ansible dry-run, idempotence and EL6 which we still support.

PR #266, opened 2020-06-26
Mongo 4.0 support for ansible - addressing issue #264

### LINKED ISSUE #264: For EL7 and U16, Install Mongo 4.0 (Ansible)
While newer recently added platforms like Ubuntu18 and CentOS8 rely on MongoDB 4,
older systems like EL7 and U16 are still on MongoDB 3.4

MongoDB v3.4 has reached its EOL in January 2020.
~Switch to MongoDB v3.6 instead~, following the StackStorm/st2docs#968
Update: Switch to MongoDB `v4.0` across all the OSes, including EL7 and U16 as we found that's supported by Mongo.

See stackstorm/st2-packages#643 for the changes needed.

amanda11:
Initial changes for 4.0. Having issues with local server, so getting a travis run performed to see if any obvious problems. 
Will change out of draft once done more verification locally.

Closes #264


amanda11:
Local testing passed - performed testing with default versions and also overriding version using the mongodb_version parameter. Ready for review.

amanda11:
> Based on discussion in `#development` team decided to use MongoDB `4.0` across all the platforms.
> Turns out previously Mongo 4 wasn't supported for EL7 and U16.
> 
> Now Mongo 4.0 is available for both platforms:
> https://docs.mongodb.com/v4.0/tutorial/install-mongodb-on-ubuntu/#platform-support
> https://docs.mongodb.com/v4.0/tutorial/install-mongodb-on-red-hat/#platform-support
> 
> This will guarantee consistency in Mongo version everywhere. Less code hacks in installers and will simplify things by a lot!

I'll update to use version 4.0 - will mark it as draft again...

amanda11:
NB. This can be simplified if we remove the redhat 6 support under the same PR. We could then go to having a single default which is the 4.0. OR we could leave it as per this PR which allows you to easily have different defaults per O/S - if we think we will go to different versions in near future on different O/S.

Another consideration is whether we want to leave all the apt-keys for the different mongo versions, if we want to allow users to choose another version of mongodb if they need to.

roles/StackStorm.mongodb/vars/redhat_6.yml
```diff
@@ -0,0 +1,7 @@
+mongo_dependencies:
+      - python-urllib3
+      - pyOpenSSL
+      - python-pyasn1
+      - python-ndg_httpsclient
+# Extract mongodb "major.minor" version
+mongodb_major_minor_version: "{{ (mongodb_version|string)[:3] }}"
```
armab, line 7: Because `mongodb_major_minor_version` is now the same across the OSes, we can move it into the shared `vars/main.yml` file.

armab: Example based on `st2` role:

https://github.com/StackStorm/ansible-st2/blob/94e3c4f37e023b2783b0de4294f2dc154d73150f/roles/StackStorm.st2/vars/main.yml

amanda11: Addressed, and I also nolonger needed redhat_6/7 variable files either - so consolidated those back into redhat.yml

armab: Thanks, looks so much cleaner! :+1:

Issue #260, opened 2020-05-29
EWC/Enterprise transition to OpenSource


armab:
Currently Ansible deployment provides functionality of installing EWC/Enterprise version.
There is a dedicated role for that: https://github.com/StackStorm/ansible-st2/tree/master/roles/StackStorm.ewc

As Extreme Networks open-sourced the EWC Enterprise (https://stackstorm.com/2020/05/27/extreme-networks-donates-ewc-to-linux-foundation/), remove the Enterprise functionality from the Ansible module. Rework the module so the settings like RBAC/LDAP could be still easily adjusted via `st2` role.

This could be done once the Enterprise functionality is integrated into core StackStorm code.

amanda11:
From discussion on the November TSC meeting then RBAC would be disabled by default to ease upgrade for community users, but would need an option to enable as a parameter (to replace the licensekey parameter)

armab:
:+1: 

Looking at the current Ansible variables config, I guess something along the lines:
```
st2_rbac:
  enable: yes
  roles: ...
  assignments: ...
```
or `st2_rbac_enable: yes`. Same for ldap.

Which will change the respective `st2.conf` settings.

winem:
I'm working on a PR now.

PR #276, opened 2020-09-29
Disable EWC install for unstable packages

armab:
Recent change to RBAC in `stackstorm/st2` master broke the EWC install for unstable packages resulting in failed builds (https://travis-ci.org/github/StackStorm/ansible-st2/jobs/730839457).

In this PR I'm just disabling the EWC installation and testing for unstable repositories by setting the `BWC_LICENSE_ENTERPRISE_UNSTABLE` ENV variable to 0.

We should however migrate the EWC role in the future st2 releases as part of https://github.com/StackStorm/ansible-st2/issues/260.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
   # Run Ansible-lint checks
   ansible-lint:
     docker:
-      - image: cytopia/ansible-lint
+      - image: yokogawa/ansible-lint
     steps:
       - checkout
       - run:
@@ -14,8 +14,8 @@ jobs:
       - run:
           name: Ansible-lint check
           command: |
-            cytopia/ansible-lint --version
-            cytopia/ansible-lint -x 204 -v roles/*/*/*.yaml roles/*/*/*.yml stackstorm.yml
+            ansible-lint --version
+            ansible-lint -x 204 -v roles/*/*/*.yaml roles/*/*/*.yml stackstorm.yml
 
   # Run YAML lint checks
   yaml-lint:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -4,22 +4,18 @@ jobs:
   # Run Ansible-lint checks
   ansible-lint:
     docker:
-      - image: williamyeh/ansible:ubuntu16.04
+      - image: yokogawa/ansible-lint
```
armab, line 7: Update: due to upstream changes in `ansible-lint` version that lint check started failing too. Fixed that as well.

stackstorm.yml
```diff
@@ -13,6 +13,6 @@
     - StackStorm.st2chatops
     - StackStorm.st2smoketests
     - role: StackStorm.ewc
-      when: ewc_license is defined and ewc_license is not none and ewc_license | length > 0
+      when: ewc_license is defined and ewc_license is not none and ewc_license | length > 1
```
amanda11, line 16: @armab How does this work? Check has changed to be that the license key is longer than 1 char, but how does that relate to:
"In this PR I'm just disabling the EWC installation and testing for unstable repositories by setting the BWC_LICENSE_ENTERPRISE_UNSTABLE ENV variable to 0."

I am presuming that the setting of the environment variable to "0" is done separately, rather than under this PR. And then this PR is making use of it, to then NOT use the ewc role when set this way.

armab: Yes, these variables are part of the TravisCI and set manually in the settings (https://travis-ci.org/github/StackStorm/ansible-st2/settings).

amanda11 (approved):
LGTM.

PR #278, opened 2020-11-26
Ansible changes for RH 7 and Python3

amanda11:
Updates for playbooks to install the libselinux-python3 module on EL 7
Updates for playbooks to check that python3-devel is available. If not available then locate the optional server repo and install python3-devel RPM from there. Needed for RHEL where python3-devel is in a disabled repo, and therefore will not get pulled in automatically as a dependency when st2 rpm is installed.

Part of StackStorm/discussions#54

### DIFF
--- a/roles/StackStorm.nginx/vars/redhat.yml
+++ b/roles/StackStorm.nginx/vars/redhat.yml
@@ -1,3 +1,4 @@
 selinux_dependencies:
   - libsemanage-python
   - libselinux-python
+  - libselinux-python3
--- a/roles/StackStorm.nginx/vars/redhat.yml
+++ b/roles/StackStorm.nginx/vars/redhat.yml
@@ -1,4 +1,3 @@
 selinux_dependencies:
   - libsemanage-python
   - libselinux-python
-  - libselinux-python3
--- a/roles/StackStorm.st2/tasks/main.yml
+++ b/roles/StackStorm.st2/tasks/main.yml
@@ -19,6 +19,7 @@
   register: _rpm_check
   args:
      warn: False
+  ignore_errors: yes
   when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '7'
   # Disable warning as yum doesn't support info
   tags: st2, skip_ansible_lint
--- a/roles/StackStorm.st2/tasks/main.yml
+++ b/roles/StackStorm.st2/tasks/main.yml
@@ -1,4 +1,53 @@
 ---
+- name: Install python3 SELinux policies
+  become: yes
+  yum:
+    name: libselinux-python3
+    state: present
+  register: _task
+  retries: 5
+  delay: 3
+  until: _task is succeeded
+  when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '7' and ansible_selinux.config_mode == "enforcing"
+  tags: st2
+
+- name: Verify python3-devel is available in enabled repo
+  become: yes
+  shell:
+    cmd: yum info python3-devel
+  changed_when: false
+  register: _rpm_check
+  args:
+     warn: False
+  when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version == '7'
+  # Disable warning as yum doesn't support info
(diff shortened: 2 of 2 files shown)


roles/StackStorm.st2/tasks/main.yml
```diff
@@ -1,4 +1,54 @@
 ---
+- name: Install python3 SELinux policies
+  become: yes
+  yum:
+    name: libselinux-python3
```
armab, line 5: Do we really need `libselinux-python3` dependency in Ansible deployment? I can't find any usage for `st2` role.

amanda11: It depends really on what python3 modules are used, ST2 itself doesn't need it, but there may be packs that import modules that need it given they are being moved from python2 -> 3, and therefore don't get the python3 selinux policies by default.

We could remove it, and then it would be up to the packs to ensure that they add the dependency if they need it.

I'm fine with either method.

armab: That dependency is already installed in `nginx` role:
https://github.com/StackStorm/ansible-st2/blob/75193b9bac1c36ef1a4ad15b5cf6433e01377c09/roles/StackStorm.nginx/vars/redhat_8.yml#L1-L3

which is then used to adjust some security restrictions that may come by default in RHEL/CentOS
https://github.com/StackStorm/ansible-st2/blob/75193b9bac1c36ef1a4ad15b5cf6433e01377c09/roles/StackStorm.nginx/tasks/nginx_redhat.yml#L81-L95


With that, we don't need it in `st2` role, but may need to update the https://github.com/StackStorm/ansible-st2/blob/master/roles/StackStorm.nginx/vars/redhat.yml

amanda11: Thanks - that's a better place to put it. Will move it to Nginx role.

roles/StackStorm.nginx/vars/redhat.yml
```diff
@@ -1,3 +1,4 @@
 selinux_dependencies:
   - libsemanage-python
   - libselinux-python
+  - libselinux-python3
```
armab, line 4: That looks now even more weird now.
If we install `libselinux-python`, why do we need `libselinux-python3` then?

Feels like we're doing something wrong.

armab: And to recap, selinux lib is used just for this:
https://github.com/StackStorm/ansible-st2/blob/d58746051d7b1aa7b574afdf1cecf506e7df34d9/roles/StackStorm.nginx/tasks/nginx_redhat.yml#L88-L95

armab: According to https://docs.ansible.com/ansible/2.5/modules/seboolean_module.html#requirements

>The below requirements are needed on the host that executes this module.
>
>    libselinux-python
>    libsemanage-python

It looks like python3 module is not needed at all in this situation.

amanda11: I've done a bit more searching... I don't believe nginx needs libselinux-python3.

I've taken it off the system, and everything is still functioning. I

However...

If we have in our code or any code that gets imported an "import selinux" then it's going to fail on the move to python3 without the include of libselinux-python3. But I've searched and none of the packs in StackStorm exchange do this explicitly (and we don't include it in the st2 repo either). 

ansible itself will still use python2, and it's a well reported one of failing if libselinux-python is not installed. So I believe we will want to keep the libselinux-python. It's also a dependency of policycoreutils-python, which is needed for the semanage call. So libselinux-python would be required.

I'm going to take libselinux-python3 out for now. And I'll do the same in the bash installer so that it's same environment.

armab: Thanks for the research, @amanda11!

This should also shed more light about the pack dependencies topic: 
https://github.com/StackStorm/st2-packages/issues/241
https://github.com/StackStorm/st2docs/issues/979#issuecomment-621767347

armab (approved):
LGTM :+1: 

Thanks for handling this corner case.

PR #279, opened 2020-12-22
Removed st2web dependency on st2 module

nmaludy:
The `StackStorm.st2web` role incorrectly declares a dependency on the `StackStorm.st2` module. This causes issues if trying to use these roles to only install individual components on a system. Example: if you only want to install `nginx` and `st2web` on a box, it also pulls in `st2`.


punkrokk (approved):

punkrokk (approved):

roles/StackStorm.st2web/defaults/main.yml
```diff
@@ -10,3 +10,6 @@ st2web_ssl_certificate_key: null
 
 # String with a custom nginx configuration file to replace st2.conf. If not provided, the default st2.conf will be used.
 st2web_nginx_config: null
+
+# StackStorm version to install. `present` to install available package, `latest` to get automatic updates or pin it to numeric version like `2.2.0`.
+st2_version: 'latest'
```
armab, line 15: If we want to split `st2web` to be independent from `st2` role, another var `st2web_version` needs to be added, instead of relying on `st2_version` from `st2` role.

Based on that, `st2_version` needs to be removed from the plays. Check `st2chatops` role for the hint, which follows the same approach.

armab (changes requested):
Thanks! The change makes sense :+1: 

I left one comment to address.

roles/StackStorm.st2web/defaults/main.yml
nmaludy: Good catch, fixing now!

README.md
```diff
@@ -48,6 +48,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2_packs`              | `[ st2 ]`     | List of packs to install. This flag does not work with a `--python3` only pack.
 | `st2_python_packages`    | `[ ]`         | List of python packages to install into the `/opt/stackstorm/st2` virtualenv. This is needed when deploying alternative auth or coordination backends which depend on Python modules to make them work.
 | **st2web**
+| `st2web_version`            | `latest`      | `st2web` version to install. `present` to install available package, `latest` to get automatic updates, or pin it to numeric version like `2.2.0`.
```
armab, line 51: Can you please also list the `st2web_revision` var in the README? This looks like oversight from the past.

roles/StackStorm.st2web/defaults/main.yml
```diff
@@ -10,3 +10,6 @@ st2web_ssl_certificate_key: null
 
 # String with a custom nginx configuration file to replace st2.conf. If not provided, the default st2.conf will be used.
 st2web_nginx_config: null
+
+# StackStorm version to install. `present` to install available package, `latest` to get automatic updates or pin it to numeric version like `2.2.0`.
+st2web_version: 'latest'
```
armab, line 15: Please move the `st2web_version` before the `st2web_revision` above to make them logically aligned as we have in other Ansible roles like `st2` and `st2chatops`.

armab:
@nmaludy Could you please provide those minor changes and so we could merge the PR?

PR #280, opened 2020-12-22
Allow datastore key to be populated from a variable.

nmaludy:
Previously, if installing StackStorm in a multi-node environment, it would generate a new datastore key on every node.

This change allows the datastore key to be specified as a variable `st2_datastore_key` then propagated to each host.

By default, this variable is `null` and will result in a random key being generated (retaining same behavior).


README.md
```diff
@@ -45,6 +45,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2_auth_username`      | `testu`       | Username used by StackStorm standalone authentication.
 | `st2_auth_password`      | `testp`       | Password used by StackStorm standalone authentication.
 | `st2_save_credentials`   | `yes`         | Save credentials for local CLI in `/root/.st2/config` file.
+| `st2_datastore_key`      | `null`        | String of the the datastore key file contents. This is useful if you're deploying StackStorm to multiple nodes and they need to share the same datastore key. By default, this variable undefined and a new random key will be generated.
```
amanda11, line 48: Suggest reword as got a "the the", and possibly missing "is", e.g. perhaps
"String for the datastore key file contents. This is useful if you're deploying StackStorm to multiple nodes and they need to share the same datastore key. By default, this variable is undefined and a new random key will be generated for each node"

punkrokk (approved):

amanda11 (commented):
LGTM - just a minor comment on the wording in the documentation.

README.md
nmaludy: Thanks, fixed!

punkrokk (approved):

amanda11 (approved):
Looks great - thanks

roles/StackStorm.st2/tasks/datastore.yml
```diff
@@ -13,6 +13,14 @@
   command: st2-generate-symmetric-crypto-key --key-path {{ st2_datastore_key_file }}
   args:
     creates: "{{ st2_datastore_key_file }}"
+  when: not st2_datastore_key
+
+- name: Copy st2 encryption key
+  become: yes
+  copy:
+    content: "{{ st2_datastore_key }}"
+    dest: "{{ st2_datastore_key_file }}"
```
armab, line 22: When the content of the file changes we'll need to notify st2 to restart the services.
According to code below it's `st2api`, but I guess that's incomplete as more StackStorm services might be using the crypto key.

armab (changes requested):

roles/StackStorm.st2/tasks/datastore.yml
nmaludy: @armab added a restart to all of the st2 services

armab: Restarting all the st2 services is impractical here. We need only specific services that are using st2 datastore key.
Can you please figure out which may use it?

PR #282, opened 2020-12-23
Support for only installing select services on a node

nmaludy:
This Ansible module supports only installing/running select services by specifying the `st2_services` variable. However, when doing so it causes issues in various cases because, for example, things like `st2ctl reload --register-all` can only be run on a `st2actionrunner` node where the pack content is available. If you're on a node just running `st2api` this command is going to fail.

This change does the following:
- Implements some simple `when` conditions for various handlers to only restart services when they are necessary
- Fixes `st2ctl` on the node to only contain the specified services in its `COMPONENTS` variable, so that `st2ctl` command still work as expected! :) 
- Fixes a small deprecation warning on the `pip` PR i did yesterday:
```
[DEPRECATION WARNING]: Invoking "pip" only once while using a loop via squash_actions is deprecated. Instead of using a loop to 
supply multiple items and specifying `name: "{{ item }}"`, please use `name: '{{ st2_python_packages }}'` and remove the loop. 
This feature will be removed from ansible-base in version 2.11. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

@armab i'm thinking about also adding some `when` conditions to the `st2smoketests` where appropriate. Could you help me figure out what the conditions should be there?

Thanks!


roles/StackStorm.st2/handlers/main.yml
```diff
@@ -8,27 +8,32 @@
 - name: reload st2
   become: yes
   command: st2ctl reload --register-all
+  when: "'st2actionrunner' in st2_services"
```
armab, line 11: Listing st2actionrunner might be very specific to implementation details in your environment.
Technically both st2api and st2sensorcontainer should have access to pack content as well.

armab (changes requested):
Thanks, @nmaludy.

As I recall, `st2_services` was added just to help with identifying services we needed to restart depending on a st2 version. This is relevant to releases when workflowengine was added and so on.

Interesting that it's now helpful for the HA compatible work. I left a couple of comments to address.
Smoke Tests seem to pass OK so no problem there.

roles/StackStorm.st2/tasks/main.yml
```diff
@@ -142,6 +141,14 @@
   loop: "{{ st2_services }}"
   tags: st2
 
+- name: Modify st2ctl to only control the given services
+  become: yes
+  lineinfile:
+    path: /opt/stackstorm/st2/bin/st2ctl
+    regexp: '^COMPONENTS='
+    line: "COMPONENTS=\"{{ st2_services | join (' ') }}\""
```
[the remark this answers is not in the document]
nmaludy: Only other thing i can think of is some way of identifying which services are supposed to be enabled on a host (thinking a config file, maybe `st2.conf`).

However, this would require changes to st2 core itself.

Relying on systemd themselves being enabled and grepping that output is unreliable as something or someone maybe have disabled a service that should be enabled.

I think this is a decent middle ground approach until we figure out if/how to support this in st2 core.

roles/StackStorm.st2/handlers/main.yml
nmaludy: The `st2api` service is "leaky" though. Given the way that this ansible setup works, you also need to install `st2api` on any host that has `st2chatops` because it is used to create the ChatOps API key.

nmaludy: could maybe do something like:
```
when: "'st2actionrunner' in st2_services and 'st2api' in st2_services"
```

roles/StackStorm.st2/tasks/main.yml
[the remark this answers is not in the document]
armab: Check out 
https://github.com/StackStorm/st2/blob/0e8ae756f30ffe2e017c64bff67830abdee7f7c9/st2common/bin/st2ctl#L20-L28

roles/StackStorm.st2/handlers/main.yml
armab: How about `st2api or st2actionrunner or st2sensorcontainer`? Technically, all these services should have access to the pack content.

PR #284, opened 2021-01-11
Updates for Ubuntu 16 and Python 3

amanda11:
Updates to scripts that if Python 3.6 is unavailable on Ubuntu 16 to ask user if they want to install the Deadsnakes PPA.

User can automatically give permission for installation from Deadsnakes PPA  by passing variable st2_u16_add_insecure_py3_ppa with value true


armab (approved):
Looks good! Thanks for codifying it :+1:

PR #285, opened 2021-01-14
Remove Chef from ansible documentation, as it is deprecated

amanda11:
Remove reference to Chef as an alternative installer, as the Chef Cookbook has been deprecated.


armab (approved):
:+1: Good find!

PR #288, opened 2021-02-09
Install ST2 pack with --force

amanda11:
Use --force when install ST2 pack to allow for fact that the "stable" ST2 is python2 on CentOS 7 and U16. But the latest st2 pack is marked as python3 only.


amanda11:
Removing, as issue will be resolved by https://github.com/StackStorm-Exchange/stackstorm-st2/pull/25

PR #289, opened 2021-03-02
Pin rabbitmq version on EL8

amanda11:
Since rabbitmq-server 3.8.13 was released, installations on EL8 failed as we install latest rabbitmq from rabbitmq packagecloud, But 3.8.13 requires version of erlang that is not in the EL8 or EPEL repositories.

This PR makes the same changes to the ansible scripts as performed by bash installer in https://github.com/StackStorm/st2-packages/pull/693


Kami (approved):
LGTM

PR #290, opened 2021-03-18
Remove EWC support & move LDAP/RBAC configuration to Stackstorm.st2

### LINKED ISSUE #260: EWC/Enterprise transition to OpenSource

winem:
This PR: 
  - removes the EWC roles
  - removes EWC/BWS from the README
  - updates links to the documentation from ewc-docs to docs.stackstorm.com
  - adds the ability to disable RBAC again (this was not possible before because the RBAC playbook was imported only when RBAC should be enabled; so disabling RBAC in the ansible variables did not actually disable it in the st2 deployment

close #260


roles/StackStorm.st2/defaults/main.yml
```diff
@@ -32,6 +32,36 @@ st2_auth_enable: yes
...
 # Password used by StackStorm standalone authentication
 st2_auth_password: testp
+
+
+# By specifying a valid configuration for LDAP,
+# (See https://docs.stackstorm.com/latest/authentication.html#ldap )
+# LDAP auth backend is setup for st2.
+# Note that you just need to provide the backend_kwargs.
+st2_ldap_enable: no
+# "enable" is not a key of st2_ldap because the defaults would be lost if any key in the dictionary is changed
```
armab, line 42: :+1:

roles/StackStorm.st2/tasks/auth-rbac.yml
```diff
@@ -0,0 +1,58 @@
...
+  ini_file:
+    dest: /etc/st2/st2.conf
+    section: rbac
+    option: enable
+    value: False
+    backup: yes
+  when: not st2_rbac_enable|bool
+  notify:
+    - restart st2api
+    - restart st2auth
```
armab, line 58: Can you please include a new line at the end of the file?
Ansible-lint has more opinions about other minor things like this.

armab (changes requested):
`+112 −740 ` - nice diff, love it! Makes things so much easier now :+1: 

Some early pointers for this PR, - looks also that the RBAC logic is missing notifying the `st2-apply-rbac-definitions --config-file /etc/st2/st2.conf` to apply the new definitions & assignments.

armab:
Related to this PR, st2docs may need an update: https://docs.stackstorm.com/install/ansible.html to reflect this new change.

It's possible to include the instructions from the history to provide the RBAC & LDAP configuration example:  
https://github.com/StackStorm/st2docs/pull/1020/files#diff-3f8f088be48f3933baeab8a1722d69bb0c9f2351670eaf2f2f4a2c9d83058e9c

roles/StackStorm.st2/defaults/main.yml
winem: Oh, you were right regarding `st2-apply-rbac-definition`. I added it now and added a task to set the rbac.backend to default (this prevented st2api from starting with rbac enabled).

winem:
> Related to this PR, st2docs may need an update: https://docs.stackstorm.com/install/ansible.html to reflect this new change.
> 
> It's possible to include the instructions from the history to provide the RBAC & LDAP configuration example:
> https://github.com/StackStorm/st2docs/pull/1020/files#diff-3f8f088be48f3933baeab8a1722d69bb0c9f2351670eaf2f2f4a2c9d83058e9c

I'll look into that over the weekend.

winem:
I provided a PR for the st2docs update: https://github.com/StackStorm/st2docs/pull/1061

.travis.yml
```diff
@@ -9,16 +9,16 @@ branches:
...
+  - DISTRO=ubuntu-16
+  - DISTRO=ubuntu-18
+  - DISTRO=centos-7
+  - DISTRO=centos-8
 
   # StackStorm 'unstable' repo check
-  - DISTRO=ubuntu-16 ST2_REPO=unstable EWC_REPO=enterprise-unstable LICENSE='BWC_LICENSE_ENTERPRISE_UNSTABLE'
-  - DISTRO=ubuntu-18 ST2_REPO=unstable EWC_REPO=enterprise-unstable LICENSE='BWC_LICENSE_ENTERPRISE_UNSTABLE'
-  - DISTRO=centos-7 ST2_REPO=unstable EWC_REPO=enterprise-unstable LICENSE='BWC_LICENSE_ENTERPRISE_UNSTABLE'
-  - DISTRO=centos-8 ST2_REPO=unstable EWC_REPO=enterprise-unstable LICENSE='BWC_LICENSE_ENTERPRISE_UNSTABLE'
```
armab, line 21: :+1: 
There's also missing https://github.com/StackStorm/ansible-st2/blob/master/.kitchen.yml file where we could remove the  license-related code.

roles/StackStorm.ewc_smoketests/tasks/main.yml
```diff
@@ -1,89 +0,0 @@
...
-  command: st2 run core.http url="https://www.google.com"
-  environment:
-    ST2_AUTH_TOKEN: "{{ st2_token_smoke_tests_user.stdout }}"
-  ignore_errors: yes
-  changed_when: no
-  register: ewc_smoke_tests_forbidden_action
-  tags:
-    - ewc-smoke-tests
-
-- name: Assert forbidden error was indeed thrown
```
armab, line 77: Looking at this file, EWC_smoketests are not bad at all.
Could we on-board them for the Community version to verify that LDAP/RBAC configuration really works?

.travis.yml
winem: Done, the hidden file slipped the grep. :smile:

roles/StackStorm.ewc_smoketests/tasks/main.yml
winem: Oh, good point. Will do that until end of this week.

winem: I'm almost done. Just need to restructure the code a bit due to a change in the recent ansible versions that prevents the flush_handlers task to be executed. Solution will be ready tomorrow.

winem: Ok, a warning led me into the wrong direction and the issue was actually something else. 

@armab I added the tests again but I had to remove the `changed_when: no` when deploying the roles and assignment template for the smoke test. Otherwise the `reload rbac` would not be triggered. I don't see an issue with it. Do you think it's fine, too? Otherwise I'll remove the notify statements and execute the handler as task regardless of any changes on these files.

roles/StackStorm.st2smoketests/defaults/main.yml
```diff
@@ -18,6 +18,6 @@ ewc_smoke_tests_rbac:
             - "runner_type_list"
 
   assignments:
-    - name: "{{ ewc_smoke_tests_user }}"
+    - name: "{{ smoke_tests_rbac_user }}"
       roles:
-        - ewc_smoke_tests_basic
+        - smoke_tests_rbac_basic
```
armab, line 23: Please include the newline at the end of this and other files where it was removed.
See https://stackoverflow.com/a/729795/4533625 for more info.

armab (commented):
Thanks for migrating the RBAC smoketests.

All looks good, however based on TravisCI jobs RBAC smoke tests are skipped. Could we enable them to make sure that the added functionality is tested in TravisCI?

roles/StackStorm.ewc_smoketests/tasks/main.yml
armab: Good to me if that works. I remember it might be needed for the Ansible idempotence checks, but don't remember for sure. As commented below let's enable the tests in CI and see the actual behavior.

roles/StackStorm.st2smoketests/defaults/main.yml
winem: Done, newlines were added. Will look into the tests tomorrow.

winem:
Enabling the RBAC tests take a bit more time than expected but I'm on it and provide them in the next days.

roles/StackStorm.st2smoketests/tasks/st2rbac_verification.yml
```diff
@@ -0,0 +1,142 @@
...
+  ini_file:
+    dest: /etc/st2/st2.conf
+    section: rbac
+    option: enable
+    value: True
+    backup: yes
+  notify:
+    - restart st2api
+    - restart st2auth
+    - reload rbac
```
armab, line 60: I guess the smoke test should really check if RBAC was configured properly and works if we set the required values in the role via vars + some specific settings for the test.

Creating directories, modifying st2.conf should remain as part of the original st2 role itself. Otherwise, we're not testing the original role behavior.

With that, I think you want to set `st2_rbac_enable: yes` that would run in CI environment only.
Similar to what was done here: https://github.com/StackStorm/ansible-st2/pull/290/files#diff-305ef63bd067b0c2e4aaae3beda8ae5ecb8d0bc9e490de22568e3e258184e50dL19-L20
and let the rbac part of the role to configure everything.

roles/StackStorm.ewc_smoketests/tasks/main.yml
```diff
@@ -1,89 +0,0 @@
...
-  fail:
-    msg: "St2 action was forbidden to run because of RBAC permissions but action still ran."
-  changed_when: no
-  when: "ewc_smoke_tests_forbidden_action.stdout.find('Forbidden') == -1"
-  tags:
-    - ewc-smoke-tests
-
-- name: Teardown test artifacts
-  import_tasks: teardown.yml
```
armab, line 86: We want to keep the teardown logic to make sure infrastructure before and after smoke tests is in the same state.

roles/StackStorm.st2smoketests/tasks/main.yml
```diff
@@ -66,3 +66,9 @@
   tags:
     - smoke-tests
     - st2chatops
+
+- name: Run RBAC tests to verify the setup
+  import_tasks: st2rbac_verification.yml
```
armab, line 71:
```suggestion
  import_tasks: st2rbac_verification.yml
  # run tests if rbac is enabled
  when: st2_rbac_enable
```

armab (changes requested):
Looks like the tests are failing on idempotence checks.

I've provided more pointers around the RBAC some tests, hope that helps.

roles/StackStorm.st2smoketests/tasks/st2rbac_verification.yml
winem: Ok, that's what I had in the first place but then I thought it's less invasive to test RBAC just as part of the smoketests. But your point is totally valid, in the first place we want to test the setup not the RBAC functionality of st2. 

I moved the code to a dedicated local branch before I switched to the "wrong" solution. So I'll switch back to the original idea. 

I'll mark the PR as WIP until we pass the tests and it's ready for review.

roles/StackStorm.st2smoketests/tasks/main.yml
winem: Done.

roles/StackStorm.ewc_smoketests/tasks/main.yml
winem: It's still there, I just moved the 3 tasks to the st2rbac_verification.yml: https://github.com/StackStorm/ansible-st2/blob/0ed1247c638801a4936ba2cdb20439fba113bdfe/roles/StackStorm.st2smoketests/tasks/st2rbac_verification.yml#L67

Let me know if we want to keep the tear down tasks in a dedicated file.

winem:
@armab this PR passes the tests now and I ran them multiple times in my own dev environment. So I hope I got everything right now.

armab (approved):
Great work! :+1: 

Thanks a lot for the diligence refactoring this.

PR #291, opened 2021-04-15
Unpin rabbitmq

amanda11:
Stop pinning version of rabbitmq on EL8.

Install latest from packagecloud, and therefore as that requires newer version of erlang, install the minimal erland provided by rabbitmq - as per instructions at https://github.com/rabbitmq/erlang-rpm

Relates to https://github.com/StackStorm/st2-packages/pull/699

### DIFF
--- a/roles/StackStorm.rabbitmq/defaults/main.yml
+++ b/roles/StackStorm.rabbitmq/defaults/main.yml
@@ -3,3 +3,5 @@ rabbitmq_plugins: []
 # To enable the management plugin (in which case you'd want at least one user tagged with administrator):
 #rabbitmq_plugins:
 #  - rabbitmq_management
+# Set to "latest" to latest version, or specify specific version
+rabbitmq_version: "latest"
--- a/roles/StackStorm.rabbitmq/defaults/main.yml
+++ b/roles/StackStorm.rabbitmq/defaults/main.yml
@@ -3,5 +3,5 @@ rabbitmq_plugins: []
 # To enable the management plugin (in which case you'd want at least one user tagged with administrator):
 #rabbitmq_plugins:
 #  - rabbitmq_management
-# Set to "latest" to latest version, or specify specific version
-rabbitmq_version: "latest"
+# Set to "present" to install latest version, or specify specific version
+rabbitmq_version: "present"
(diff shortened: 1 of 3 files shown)


Kami:
I personally still pinning to a specific is a good idea / best practice to avoid surprises due to major version changes or similar and ensure reproducibility. 

But otherwise change looks good to me :+1:

roles/StackStorm.rabbitmq/defaults/main.yml
```diff
@@ -3,3 +3,5 @@ rabbitmq_plugins: []
 # To enable the management plugin (in which case you'd want at least one user tagged with administrator):
 #rabbitmq_plugins:
 #  - rabbitmq_management
+# Set to "latest" to latest version, or specify specific version
+rabbitmq_version: "latest"
```
armab, line 7: I think what @Kami wanted to say is that expectations from the 3rd party software that stackstorm relies on is to use `present` which will install the version once and won't update it automatically as it happens with the `latest`.

Looks like we're following the same with the `mongodb`, `nginx` and `nodejs` roles.

amanda11: Aha - makes sense - I'll update.

amanda11: @armab  Have updated variable to use "present", it was using present in the tasks - but name could have been confusing - so went back to naming it "present".
Let me know if that isn't what we wanted.

Kami:
@amanda11 

> If we want to pin version we should probably do it across the board - on all installers, and define version for each OS.
I've left in the functionality so that you can still the pin the version if you desired - as I figured that was worth keeping.

Yeah, in this case it's fine for now.

> @Kami What about if I create a generic issue under ST2 to add in support to pin rabbitmq version, as if we do it we should consider it for all OS and all installers, and make sure that we keep all install methods and docs etc using same version? As I think its a bigger issue than just this EL8 ansible problem.

Yeah, imo that would be ideal long term scenario.

Technically, we already kinda (unintentionally) do that for some software - e.g. MongoDB which uses different apt repo per major version. And for RabbitMQ as well (to some extent) since we rely on default upstream versions which usually don't ship all the most recent major versions (which is now not true for the RabbitMQ apt repo anymore).

Kami:
st2vagrantdev change (confirmed working locally) - https://github.com/StackStorm/st2vagrantdev/pull/10

armab (approved):
:+1:

Issue #292, opened 2021-04-19
EPEL repo install failed with: Failed to validate GPG signature for epel-release-8-10.el8.noarch


zen:
Hi,

After miserably failing with st2-docker (DNS resolution issue with containers in bridged networks) I attempted to install ST2 through ansible playbooks. 

ansible --version
ansible 2.10.7

Environment is AWS deployed official Centos 8 image, python 3 installed on host. Configuration:

```
- name: Instal Stackstorm
  hosts: stackstorm
  become: yes
  # ST2 does not directly support AWS Linux hence:
  remote_user: centos
  vars:
    st2_system_user: st2-user
    st2_packs: [ aws, consul, ansible, github, prometheus, slack, terraform, vault ]

  roles:
    - StackStorm.mongodb
    - StackStorm.rabbitmq
    - StackStorm.st2repo
    - StackStorm.st2
    - StackStorm.nginx
    - StackStorm.st2web
    - StackStorm.nodejs
    - StackStorm.st2chatops
    - StackStorm.st2smoketests
```

Result is as follows:

```
PLAYBOOK: stackstorm.yaml ****************************************************************************************************************************************************************************************
1 plays in stackstorm.yaml

PLAY [Instal Stackstorm] *****************************************************************************************************************************************************************************************

TASK [Gathering Facts] *******************************************************************************************************************************************************************************************
task path: /Users/zen/projects/work/al/infrastructure/ansible/ec2-setup/stackstorm.yaml:9
ok: [stackstorm-sing]
META: ran handlers

TASK [StackStorm.epel : Check if EPEL is installed] **************************************************************************************************************************************************************
task path: /Users/zen/.ansible/roles/stackstorm.stackstorm/roles/StackStorm.epel/tasks/main.yml:2
ok: [stackstorm-sing] => {"changed": false, "stat": {"exists": false}}

TASK [StackStorm.epel : Install EPEL repo] ***********************************************************************************************************************************************************************
task path: /Users/zen/.ansible/roles/stackstorm.stackstorm/roles/StackStorm.epel/tasks/main.yml:9
FAILED - RETRYING: Install EPEL repo (5 retries left).
FAILED - RETRYING: Install EPEL repo (4 retries left).
FAILED - RETRYING: Install EPEL repo (3 retries left).
FAILED - RETRYING: Install EPEL repo (2 retries left).
FAILED - RETRYING: Install EPEL repo (1 retries left).
[...]

PR #294, opened 2021-04-20
Download GPG key explicitly

amanda11:
Closes https://github.com/StackStorm/ansible-st2/issues/292

Not sure why this has started to fail recently. But downloaded the key first, ensures that it downloads successfully.

On bash installers as we use -y then it will install without issue


armab (approved):
:+1:

PR #295, opened 2021-04-20
Fix problem installing specific version of ST2

### LINKED ISSUE #150: 'st2_version' Overhaul
The goal is consistent functionality incl. naming when it comes to specifying Stackstorm version to install to allow users achieve what they want: major/minor/path version, auto-update, no-autoupdate, latest revision pin, etc.

- [x] Fix version install syntax for `apt`/`yum` (done in https://github.com/StackStorm/ansible-st2/pull/148/commits/dfb5588d57d2cfc390e73a1f226ffd8514d0ce77)
- [ ] Consider deprecating `st2_revision` since it doesn't work everywhere
  - `apt install st2=2.2.1-5` (revision is required)
  - `yum install st2-2.2.1` (revision could be omitted)
- [ ] Consistency: not every role contains `st2_revision`. Use everywhere or nowhere.
- [ ] Naming overhaul for `st2_version`: `null`, ``, `latest`, `present`, `*`
- [ ] Consider changing `st2_version` defaults to NO auto-update, as prod-friendly setting
- [ ] Research solution for https://github.com/StackStorm/ansible-st2/issues/111
- [ ] Better README with version install examples & tricks & describing diff apt/yum behavior to allow users achieve what they want (stable, patch, auto-update, no-autoupdate, latest revision pin)

amanda11:
Currently the playbooks do not work if you want to specify a particular version of ST2.

The problems were:

- wrong check on using = or - as separately, as logic was if yum else do ... But newer EL are dnf - and so were using the apt separator
- st2_version was used for both st2 and st2web packages. And if specified always had to specify st2_revision. This only works if the same revision exists for both packages, but this isn't necessarily the case e.g. 3.4.0 has 3.4.0-1 for st2, and 3.4.0-3 on st2web
- On EL systems then there is no need to specify revisions, if use 3.4.0 it would give the latest version. But on Debian the revision needs to be specified
- st2chatops was insatlled differntly with a single st2chatops_version that could be used as with or without revision

These issues were raised in https://github.com/StackStorm/ansible-st2/issues/150
Closes #150

Updated the playboosk to use the current revision separate for Debian and EL.
Updated the playbooks to use a consistent approach for revisioning, using st2_version, st2chatops_version and st2web_version. This allows the user to either specify with revision for EL and Debian, or they can omit revision on EL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,10 +48,10 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2_python_packages`    | `[ ]`         | List of python packages to install into the `/opt/stackstorm/st2` virtualenv. This is needed when deploying alternative auth or coordination backends which depend on Python modules to make them work.
 | `st2_u16_add_insecure_py3_ppa`	| `false`     | Whether permission is granted to install the deadsnakes Python3.6 PPA for Ubuntu 16. 
 | **st2web**
+| `st2web_version`     | `latest`      | st2web version to install. `present` to install available package, `latest` to get automatic updates, or pin it to numeric version like `2.2.0` or with revision like `2.2.0-1`.
 | `st2web_ssl_certificate`     | `null` | String with custom SSL certificate (`.crt`). If not provided, self-signed certificate will be generated.
 | `st2web_ssl_certificate_key` | `null` | String with custom SSL certificate secret key (`.key`). If not provided, self-signed certificate will be generated.
 | `st2web_nginx_config`     | `null` | String with a custom nginx configuration file (`st2.conf`). If not provided, the default st2.conf will be used.
-| `st2web_version`     | `latest`      | st2web version to install. `present` to install available package, `latest` to get automatic updates, or pin it to numeric version like `2.2.0` or with revision like `2.2.0-1`.
 | **ewc**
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -52,6 +51,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2web_ssl_certificate`     | `null` | String with custom SSL certificate (`.crt`). If not provided, self-signed certificate will be generated.
 | `st2web_ssl_certificate_key` | `null` | String with custom SSL certificate secret key (`.key`). If not provided, self-signed certificate will be generated.
 | `st2web_nginx_config`     | `null` | String with a custom nginx configuration file (`st2.conf`). If not provided, the default st2.conf will be used.
+| `st2web_version`     | `latest`      | st2web version to install. `present` to install available package, `latest` to get automatic updates, or pin it to numeric version like `2.2.0` or with revision like `2.2.0-1`.
```
armab, line 54: For better visibility and similar to other roles, perhaps it makes sense to move the `st2web_version` var description higher in the list.

armab (approved):
:+1: 

Makes the versioning simpler.

PR #296, opened 2021-04-27
Updates to install redis, so can be used as coordination backend

amanda11:
Updates to make redis the default backend.

Introduced new role for redis, so that it can be remove or  replaced if different coordination backend required.

ST2 unstable (and therefore 3.5) will come with libraries for redis client, and therefore redis can currently be enabled by setting st2_config ansible variable to define the coordination url.

Once bash installer also includes redis configuration, then we can make the default st2.conf  specify redis url as the default


amanda11:
Not sure why tests on Ubuntu are failing when st2chatops needs the st2apikey. Have run the playbook successfully on an Ubuntu Bionic fresh VM and all passed.

roles/StackStorm.redis/tasks/redis_redhat.yml
```diff
@@ -0,0 +1,19 @@
+---
+- name: Install redis
+  become: yes
+  package:
+    name: redis
```
armab, line 5: I see the package naming is different between the el/deb.
If that's the only case, maybe we could do it via vars instead and keep the package installation in a single main file?
https://github.com/StackStorm/ansible-st2/blob/b15ea27511ccf579809c2d5ceb0fe3e6f6fbfda3/roles/StackStorm.mongodb/tasks/main.yml#L2-L6

roles/StackStorm.st2/defaults/main.yml
```diff
@@ -46,3 +46,7 @@ st2_python_packages: []
 #Alternatively the playbooks will  try to add python3.6 from the 3rd party 'deadsnakes' repository: https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa.
 #Only set to true, if you are aware of the support and security risks associated with using unofficial 3rd party PPA repository, and you understand that StackStorm does NOT provide ANY support for python3.6 packages on Ubuntu 16.04.
 st2_u16_add_insecure_py3_ppa: false
+
+# If defined, configure coordination_url for st2 to use. By default this will
+# be redis on localhost. 
+st2_coordination_url: redis://127.0.0.1:6379
```
armab, line 52: What if we define this default via the existing `st2_config`?

Better options if we're making the coordination enabled by default since the next st2 release, I believe it should be just part of the original `st2.conf` and so Ansible coordination string would be not needed.

Similar to what we already do for Mongo and RabbitMQ:
https://github.com/StackStorm/st2/blob/258e0440da5c21a2cdcb836749ca0dbf74c5d292/conf/st2.package.conf#L75-L76

amanda11:
Found problem, with redis on Ubuntu. redis attempts to bind on both IPv4 and IPv6, but it fails to bind in our docker containers. Changing the bind to only be 127.0.0.1 works.
Will work on fix for playbooks.

amanda11:
In case of stable runs then we don't want coordination url to redis, as ST2 3.4.1 won't have relevant python libraries configured. 

If the Travis run passes on the unstable with the latest changes, then I will take out the defaults for coordination url in st2_config parameter. So it won't use redis coordination url unless it's the default in st2.conf in the package.

This should then pass for unstable and stable, but ST2 won't actually be using redis. But we'll know the playbooks are setup to configure redis for when url is set to redis.

Can't change the deafult in st2.conf until we have the bash installer changes under https://github.com/StackStorm/st2-packages/pull/698 merged. But once that is done we can make the co-ordination url default to redis.

armab (approved):
Thanks, looks great! 

That's even better :+1:, meaning [coordination] will automagically work once the Redis becomes the default backend in the upstream `st2.conf`.

PR #298, opened 2021-06-08
Drop U16 and add U20 support

winem:
This PR drops the support for Ubuntu 16 (xenial) and adds support for Ubuntu 20 (focal). 

In addition this PR also adds meta/main.yml files for the redis and smoketests role which weren't there yet. This is rather a matter of consistency than of functionality but it seems a good chance to get this done.

It also removes support for MongoDB-Versions that are already EoL as there are 3.2 and 3.4.

Tests with st2repo_name = unstable were successful.


armab:
Looks like we'll also need to add U 20 to CI, per https://travis-ci.org/github/StackStorm/ansible-st2/builds/773942784

roles/StackStorm.redis/meta/main.yml
```diff
@@ -0,0 +1,27 @@
...
+    - system
+    - stackstorm
+    - st2
+    - automation
+    - remediation
+    - devops
+dependencies:
+  - role: StackStorm.epel
+    when: ansible_facts.os_family == 'RedHat'
+  - role: StackStorm.st2repo
```
cognifloyd, line 27: Why does the `StackStorm.redis` role need the st2repo?

roles/StackStorm.st2/tasks/auth.yml
```diff
@@ -1,24 +1,23 @@
+- name: Include OS-specific variables
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ ansible_os_family | lower }}_{{ ansible_distribution_major_version }}.yml"
+    - "{{ ansible_os_family | lower }}.yml"
```
cognifloyd, line 5:
```suggestion
    - "{{ ansible_facts.os_family | lower }}_{{ ansible_facts.distribution_major_version }}.yml"
    - "{{ ansible_facts.os_family | lower }}.yml"
```

roles/StackStorm.mongodb/vars/debian_20.yml
```diff
@@ -0,0 +1,7 @@
+# Use the following URL to find the key: https://www.mongodb.org/static/pgp/server-{{ mongodb_major_minor_version }}.asc
+mongodb_apt_keys:
+  "4.2": "3F75C80EE56C41F4E49515B62ACD4032058F8B6B"
+  "4.4": "20691EEC35216C63CAF66CE1656408E390CFB1F5"
```
cognifloyd, line 4: Duplicating these keys is not ideal. Is the idea here to prevent people from trying to install 3.6 or 4.0 on debian 20?

cognifloyd (changes requested):
I have few minor changes and questions.

roles/StackStorm.mongodb/vars/debian.yml
```diff
@@ -1,6 +1,6 @@
 # Use the following URL to find the key: https://www.mongodb.org/static/pgp/server-{{ mongodb_major_minor_version }}.asc
 mongodb_apt_keys:
-  "3.2": "42F3E95A2C4F08279C4960ADD68FA50FEA312927"
-  "3.4": "0C49F3730359A14518585931BC711F9BA15703C6"
   "3.6": "2930ADAE8CAF5059EE73BB4B58712A2291FA4AD5"
   "4.0": "9DA31620334BD75D9DCB49F368818C72E52529D4"
+  "4.2": "3F75C80EE56C41F4E49515B62ACD4032058F8B6B"
```
amanda11, line 5: I think as we haven't yet tested with 4.2, then I'd be tempted to only add 4.4 for now.
Add 4.2 when we confirm if we want to support it as part of 3.6

amanda11 (changes requested):
As well as the comment on mongo 4.2, I think we also need updates on:

- .kitchen.yml - has ubuntu 16 still
- .travis.yml
-

roles/StackStorm.mongodb/vars/debian_20.yml
winem: I don't think that it would prevent people from installing an older version. There is no package for mongodb 3.6 or 4.0 on focal. So trying to install mongodb 4.0 for example would fail while adding the key as such would instead work. 

So the keys aren't needed for U20 and I thought duplication is acceptable as it keeps the code clean and the task to retrieve the variables does not need to be changed: 
```
---
- name: Retrieve variables
  include_vars: "{{ item }}"
  with_first_found:
    - "{{ ansible_os_family | lower }}_{{ ansible_distribution_major_version }}.yml"
    - "{{ ansible_os_family | lower }}.yml"
  tags: [databases, mongodb]
``` 

I think I'm going to add another task that loads the debian.yml that has all the apt keys and the debian_20.yml only if it's a U20 setup to override the mongodb version. Will push it after a test.

winem: Done with https://github.com/StackStorm/ansible-st2/pull/298/commits/df3e85cedcca5a63f94dddcfac8cc28b89d0652d

roles/StackStorm.redis/meta/main.yml
winem: You are right, there is no technical dependency between both. Removed it from the meta file now.

roles/StackStorm.mongodb/vars/debian.yml
winem: Ok, fine for me. I removed it.

winem:
Will update the ci files (travis, kitchen, etc.) till tomorrow evening.

cognifloyd:
`mongodb_version` needs to change for debian 20. You added an extra vars file and tasks to load that. I think it would be cleaner to consolidate the version choice in the defaults file, which looks like:

https://github.com/StackStorm/ansible-st2/blob/94e3c4f37e023b2783b0de4294f2dc154d73150f/roles/StackStorm.mongodb/defaults/main.yml#L2-L3

But you could do this:
```diff
 # MongoDB default version
-mongodb_version: "4.0" 
+mongodb_version: "{% if ansible_facts.os_family == 'Debian' and ansible_facts.distribution_major_version == '20' %}4.4{% else %}4.0{% endif %}" 
```

roles/StackStorm.mongodb/tasks/main.yml
```diff
@@ -1,9 +1,20 @@
...
+
+- name: Retrieve variables (Debian)
+  include_vars: debian.yml
+  when: ansible_facts.os_family == 'Debian'
+  tags: [databases, mongodb]
+
+- name: Retrieve variables (Debian)
+  include_vars: debian_20.yml
+  when: ansible_facts.os_family == 'Debian' and ansible_facts.distribution_major_version == '20'
   tags: [databases, mongodb]
```
armab, line 18: This 3-tasks construct doesn't look good indeed.
Avoiding os-specific `when` condition is exactly the purpose of the 1st task `with_first_found` which is a known best practice.

winem: That's true and imho it's actually an anti-pattern to override defaults via role variables. This has a high potential to cause failures and confuse maintainers. Now I switched to the solution suggested by @cognifloyd which seems to be the most elegant one.

winem:
> `mongodb_version` needs to change for debian 20. You added an extra vars file and tasks to load that. I think it would be cleaner to consolidate the version choice in the defaults file, which looks like:
> 
> https://github.com/StackStorm/ansible-st2/blob/94e3c4f37e023b2783b0de4294f2dc154d73150f/roles/StackStorm.mongodb/defaults/main.yml#L2-L3
> 
> But you could do this:
> 
> ```diff
>  # MongoDB default version
> -mongodb_version: "4.0" 
> +mongodb_version: "{% if ansible_facts.os_family == 'Debian' and ansible_facts.distribution_major_version == '20' %}4.4{% else %}4.0{% endif %}" 
> ```

I did it like that now. Guess that's the smartest solution to the issue and does not break with any common pattern or best practices.

amanda11 (approved):
LGTM - but wait for those others that gave review comments to give their approval.

cognifloyd (approved):
LGTM

.kitchen.yml
```diff
@@ -22,18 +22,18 @@ provisioner:
...
       run_command: /sbin/init
       volume:
         - /sys/fs/cgroup:/sys/fs/cgroup:ro
-  # Ubuntu Bionic with Systemd
-  - name: ubuntu-18.04
+  # Ubuntu Focal with Systemd
+  - name: ubuntu-20.04
     driver_config:
-      image: stackstorm/packagingtest:bionic-systemd
+      image: stackstorm/packagingtest:xenial-systemd
```
armab, line 36: You probably meant `focal` here:
```suggestion
      image: stackstorm/packagingtest:focal-systemd
```

.travis.yml
```diff
@@ -9,14 +9,14 @@ branches:
 
 env:
   # default is stable repo
-  - DISTRO=ubuntu-16
   - DISTRO=ubuntu-18
+  - DISTRO=ubuntu-20
```
amanda11, line 13: Need to remove ubuntu-20 from here. We can't run it as a stable run yet until 3.5 is released. 
So we need to do the changes for ansible-st2 in 2 stages.
1) add it to unstable before release - under this PR
2) add it to stable after release - separate PR

amanda11 (changes requested):

.travis.yml
armab: I think we'll just have to wait for a release to happen before merging this PR.

Otherwise, someone using the fresh Ansible installer will have the same issues with U16/U20.

roles/StackStorm.mongodb/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
 # MongoDB default version
-mongodb_version: "4.0"
+mongodb_version: "{% if ansible_facts.os_family == 'Debian' and ansible_facts.distribution_major_version == '20' %}4.4{% else %}4.0{% endif %}"
```
rush-skills, line 3: To decide between 4.0 and 4.4 mongo version, wouldn't it be a better idea to use st2 version > 3.4 rather than only on Ubuntu 20?

amanda11: @rushskills Stackstorm 3.5 still uses mongo 4.0 for all OSes EXCEPT for Ubuntu 30

armab (approved):
:+1: 
Following the #Ansible community messages, we want to merge it, considering the 3.5.0 release is ready.

Just migrated repo to the new TravisCI.com to get the build finished.

armab:
All green, great work everyone!

PR #299, opened 2021-07-04
Add support for nodejs 14 and list the variable on the README

winem:
There were different ways to realize the node 14 support but this one is the only one that's not breaking. 

Changing just the default would break deployments of older releases even though we may mention it on the README. 2nd challenge was that a simple comparison with the ansible version test is not able to work with `latest` as st2_version. That's why we use an if-elsif-else and not just an if-else clause here.

### DIFF
--- a/roles/StackStorm.nodejs/defaults/main.yml
+++ b/roles/StackStorm.nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "{% if st2_version == 'latest' %}14{% elif st2_version is version('3.5.0', '<') %}10{% else %}14{% endif %}"
+nodejs_major_version: "{% if st2_version == 'latest' or st2_version is version('3.5.0', '<') %}14{% else %}10{% endif %}"
--- a/roles/StackStorm.nodejs/defaults/main.yml
+++ b/roles/StackStorm.nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "10"
+nodejs_major_version: "{% if st2_version == 'latest' %}14{% elif st2_version is version('3.5.0', '<') %}10{% else %}14{% endif %}"
--- a/roles/StackStorm.nodejs/defaults/main.yml
+++ b/roles/StackStorm.nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "{% if st2_version == 'latest' or st2_version is version('3.5.0', '<') %}14{% else %}10{% endif %}"
+nodejs_major_version: "{% if st2chatops_version == 'latest' or st2chatops_version is version('3.5.0', '>=') %}14{% else %}10{% endif %}"
--- a/roles/StackStorm.nodejs/defaults/main.yml
+++ b/roles/StackStorm.nodejs/defaults/main.yml
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "{% if st2chatops_version == 'latest' or st2chatops_version is version('3.5.0', '>=') %}14{% else %}10{% endif %}"
+nodejs_major_version: "14"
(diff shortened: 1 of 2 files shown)


roles/StackStorm.nodejs/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "10"
+nodejs_major_version: "{% if st2_version == 'latest' %}14{% elif st2_version is version('3.5.0', '<') %}10{% else %}14{% endif %}"
```
cognifloyd, line 3:
```suggestion
nodejs_major_version: "{% if st2_version == 'latest' or st2_version is version('3.5.0', '>=') %}14{% else %}10`{% endif %}"
```
What about something like this?

amanda11 (commented):
Looks good, couple of comments.

Does this cope if you specify st2_version as 3.5.0-1 - as for Ubuntu we have to specify the full version of package, whereas for El we can just use 3.5.0.

Also probably best to compare with st2chatops _version instead of st2_version.

roles/StackStorm.nodejs/defaults/main.yml
winem: Good catch! I tried it with a different order and ran into the issue that version() does not expect a string. Changed it as per your suggestion.

winem:
> Looks good, couple of comments.
> 
> Does this cope if you specify st2_version as 3.5.0-1 - as for Ubuntu we have to specify the full version of package, whereas for El we can just use 3.5.0.
> 
> Also probably best to compare with st2chatops _version instead of st2_version.

I'll mark the PR as draft for now and do another round of tests with the st2chatops_version later tonight. 

Is there a reason why `st2chatops_version` does not reference `st2_version` as default?

amanda11:
> > Looks good, couple of comments.
> > Does this cope if you specify st2_version as 3.5.0-1 - as for Ubuntu we have to specify the full version of package, whereas for El we can just use 3.5.0.
> > Also probably best to compare with st2chatops _version instead of st2_version.
> 
> Yes, it does. I tested it with `3.5.0-1`, `3.4.1-2` and `latest`.
> 
> I'll mark the PR as draft for now and do another round of tests with the st2chatops_version later tonight.
> 
> Is there a reason why `st2chatops_version` does not reference `st2_version` as default?

st2chatops_version and st2_version aren't necessarily the same on ubuntu. e.g. for 3.4.1 - then st2_version needs to be 3.4.1-2 and st2chatops_version needs to be 3.4.1-1. 
(If it was just EL then everything woulsd be nice and simple and just one variable....  - I also think the roles are generally designed to be independant, but if you find other cases where the st2chatops role picks up st2_version, then I don't have an issue with st2chatops_version defaulting to st2_version. Though you do need to make sure you check the package versions, so if we end up defaulting them I can see people not bothering to check, and then accidentally installing 3.4.1-1 st2 - which isn't the latest...

winem:
I wasn't aware that both are released independent from each other. So that's what I call a good reason to have the vars being independent as well. Thanks!

Tests went fine so this PR is open for final reviews.

roles/StackStorm.nodejs/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
 # nodejs version to install
-nodejs_major_version: "10"
+nodejs_major_version: "{% if st2chatops_version == 'latest' or st2chatops_version is version('3.5.0', '>=') %}14{% else %}10{% endif %}"
```
armab, line 3: What if the user sets the `st2chatops_version=present`?
This is possible for both scenarios when the older st2 version and newer st2 version is installed.

armab: In Ansible deployment we normally don't provide any migration, support, or handling for the previous st2 versions. And so it means we just support the latest st2 version, allowing users to pin the Ansible roles to the previous version if they want it to work with the older st2.

With that, I'd just recommend removing the conditional and set it to `14` by default.

armab, line 3: In other words,
```suggestion
nodejs_major_version: "14"
```
is all we need

winem: Thanks for the input. That's good to know and will be helpful for my future contributions. Did it as suggested.

winem: Done. And I updated the README. Hope it's fine now.

armab (approved):
Thanks for the update!

PR #302, opened 2021-10-21
Install latest rabbitmq on all OSes

amanda11:
Update the ansible installer to install latest rabbitmq from PackageCloud, rather than version bundled in OS.


roles/StackStorm.rabbitmq/tasks/rabbitmq_redhat.yml
```diff
@@ -0,0 +1,59 @@
+---
+
+- name: RabbitMQ on RHEL8 requires module(perl:5.26)
```
armab, line 3: Considering we moved to official RabbitMQ, is this requirement still a thing?
I see it's from the previous Ansible role.

amanda11: @armab I took it out and install was fine, so no-longer required.

roles/StackStorm.rabbitmq/tasks/rabbitmq_redhat.yml
```diff
@@ -0,0 +1,46 @@
...
+    description: rabbitmq_erlang
+    baseurl: https://packagecloud.io/rabbitmq/erlang/el/{{ ansible_facts.distribution_major_version }}/$basearch
+    repo_gpgcheck: yes
+    gpgkey: "https://packagecloud.io/rabbitmq/erlang/gpgkey"
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
+    metadata_expire: 300
+    gpgcheck: no
+    enabled: yes
+    sslverify: yes
+  when: ansible_facts.os_family == 'RedHat'
```
armab, line 17: Considering the file is named `rabbitmq_redhat.yml` which is included conditionally depending on OS,
I think this line could be removed.
```suggestion
```

roles/StackStorm.rabbitmq/tasks/rabbitmq_redhat.yml
```diff
@@ -0,0 +1,46 @@
...
+    description: rabbitmq-server
+    baseurl: https://packagecloud.io/rabbitmq/rabbitmq-server/el/{{ ansible_facts.distribution_major_version }}/$basearch
+    repo_gpgcheck: yes
+    gpgkey: "https://packagecloud.io/rabbitmq/rabbitmq-server/gpgkey"
+    sslcacert: /etc/pki/tls/certs/ca-bundle.crt
+    metadata_expire: 300
+    gpgcheck: no
+    enabled: yes
+    sslverify: yes
+  when: ansible_facts.os_family == 'RedHat'
```
armab, line 34:
```suggestion
```

amanda11: @armab  Good spot - will remove!

PR #303, opened 2021-11-09
Move tests from Travis to GitHub Actions

### LINKED ISSUE #297: Migrate to Github Actions
This repository is using TravisCI for testing its Ansible run in Docker.
https://travis-ci.org/github/StackStorm/ansible-st2/builds/770421377

And while the Travis still works, we'll need to migrate to GH Actions as Travis will be probably decommissioned later.
Kind of our technical debt.

amanda11:
Resolves #297


winem (approved):
Thanks!

armab (commented):
Nice one!

The only thing I noticed, - the build here is happening twice.
image

Would be nice to have it run once for PRs, otherwise, it'll generate lots of noise.
Another point, parallel usage of Slack token in several jobs will also fail on rate-limiting. So minimizing the jobs would be helpful.

armab (approved):
I can only think of the build badge update in the `README.md`

Besides that, looks great :+1:

amanda11:
> I can only think of the build badge update in the `README.md`
> 
> Besides that, looks great 👍

Good point - thanks for pointing that out!

PR #310, opened 2022-05-03
Adding support for custom SSL by path

### LINKED ISSUE #309: Support for custom SSL certificate and key (by path)
There are options for providing `st2web_ssl_certificate` and `st2web_ssl_certificate_key`, which work just fine. It would be nice to alternatively provide paths to these files instead (versus the string contents). I'm currently working on a deployment which leverages a separate role to generate the certificate and key and it'd be convenient to be able to reference their paths (versus fetching them into strings to feed to the st2 roles).

mamercad:
Fixes #309.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,8 @@ Below is the list of variables you can redefine in your playbook to customize st
 | `st2web_version`     | `latest`      | st2web version to install. `present` to install available package, `latest` to get automatic updates, or pin it to numeric version like `2.2.0` or with revision like `2.2.0-1`.
 | `st2web_ssl_certificate`     | `null` | String with custom SSL certificate (`.crt`). If not provided, self-signed certificate will be generated.
 | `st2web_ssl_certificate_key` | `null` | String with custom SSL certificate secret key (`.key`). If not provided, self-signed certificate will be generated.
+| `st2web_ssl_certificate_path`     | `null` | Path to custom SSL certificate (mutally-exclusive with `st2web_ssl_certificate` and `st2web_ssl_certificate_key`).
+| `st2web_ssl_certificate_key_path` | `null` | Path to custom SSL certificate secret key (mutally-exclusive with `st2web_ssl_certificate` and `st2web_ssl_certificate_key`).
 | `st2web_nginx_config`     | `null` | String with a custom nginx configuration file (`st2.conf`). If not provided, the default st2.conf will be used.
 | **st2chatops**
 | `nodejs_major_version`   | `14`          | The default fits st2chatops version >= 3.5.0.
--- a/roles/StackStorm.st2web/defaults/main.yml
+++ b/roles/StackStorm.st2web/defaults/main.yml
@@ -10,5 +10,11 @@ st2web_ssl_certificate: null
 # String with custom SSL certificate private key. If not provided, self-signed certificate will be generated.
 st2web_ssl_certificate_key: null
 
+# Path to custom SSL certificate (mutally-exclusive with st2web_ssl_certificate and  st2web_ssl_certificate_key).
+st2web_ssl_certificate_path: null
+
+# Path to custom SSL certificate private key (mutally-exclusive with st2web_ssl_certificate and  st2web_ssl_certificate_key).
+st2web_ssl_certificate_key_path: null
+
 # String with a custom nginx configuration file to replace st2.conf. If not provided, the default st2.conf will be used.
 st2web_nginx_config: null
--- a/roles/StackStorm.st2web/tasks/certificate.yml
+++ b/roles/StackStorm.st2web/tasks/certificate.yml
@@ -5,6 +5,24 @@
   # no XOR in Yaml
   when: (st2web_ssl_certificate and not st2web_ssl_certificate_key) or (not st2web_ssl_certificate and st2web_ssl_certificate_key)
 
+- name: Verify if custom SSL certificate - by path - was correctly specified
+  fail:
+    msg: "When using custom certificate by path, both 'st2web_ssl_certificate_path' and 'st2web_ssl_certificate_key_path' must be provided"
+  # no XOR in Yaml
+  when: (st2web_ssl_certificate_path and not st2web_ssl_certificate_key_path) or (not st2web_ssl_certificate_path and st2web_ssl_certificate_key_path)
+
+- name: Verify if custom SSL certificate by string - and - by path (this is not supported)
+  fail:
+    msg: "When using custom certificate, cannot use both 'st2web_ssl_certificate' and 'st2web_ssl_certificate_path'"
+  # no XOR in Yaml
+  when: st2web_ssl_certificate and st2web_ssl_certificate_path
+
+- name: Verify if custom SSL certificate key by string - and - by path (this is not supported)
+  fail:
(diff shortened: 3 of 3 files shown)


armab (commented):
Do we introduce 2 new variables just to do the same as the original logic has, but with different input (via remote file vs via text)?

In https://github.com/StackStorm/ansible-st2/issues/309 you mentioned about managing the certificates in another role.

Could you write the needed certificate contents directly to the `/etc/ssl/st2/st2.crt` and `/etc/ssl/st2/st2.key` from that role?
This way you won't need to do anything else in the stackstorm role, it'll just pick it up.

mamercad:
> Do we introduce 2 new variables just to do the same as the original logic has, but with different input (via remote file vs via text)?
> 
> In #309 you mentioned about managing the certificates in another role.
> 
> Could you write the needed certificate contents directly to the `/etc/ssl/st2/st2.crt` and `/etc/ssl/st2/st2.key` from that role? This way you won't need to do anything else in the stackstorm role, it'll just pick it up.

Ah, that's interesting, I'll look into this.

mamercad:
> This way you won't need to do anything else in the stackstorm role, it'll just pick it up.

I don't think it'll just work ... based on [this](https://github.com/StackStorm/ansible-st2/blob/70300aab54a904295dbab21f50cef5486eaba1a1/roles/StackStorm.st2web/tasks/certificate.yml#L20), both `st2web_ssl_certificate` and `st2web_ssl_certificate_key` need to have the "value" of the certificate and the key (notice the combination of the `copy` module and the `content` parameter).

The external role that I'm using is doing LetsEncrypt ... I generate a private key and a CSR, submit it, and then fetch back the certificate and chain (storing all of these files on disk) after the LE stuff happens.

Am I missing something -- thoughts @armab?

mamercad:
> > This way you won't need to do anything else in the stackstorm role, it'll just pick it up.
> 
> I don't think it'll just work ... based on [this](https://github.com/StackStorm/ansible-st2/blob/70300aab54a904295dbab21f50cef5486eaba1a1/roles/StackStorm.st2web/tasks/certificate.yml#L20), both `st2web_ssl_certificate` and `st2web_ssl_certificate_key` need to have the "value" of the certificate and the key (notice the combination of the `copy` module and the `content` parameter).
> 
> The external role that I'm using is doing LetsEncrypt ... I generate a private key and a CSR, submit it, and then fetch back the certificate and chain (storing all of these files on disk) after the LE stuff happens.
> 
> Am I missing something -- thoughts @armab?

Oh, you mean, use my role after the StackStorm roles/plays and just clobber `/etc/ssl/st2/st2.crt` and `/etc/ssl/st2/st2.key` (followed by restarting `nginx`)? I suppose that'd work ... that's not very nice though is it?

armab:
Perhaps `cp` them on a remote host or similar as a prep task before running the st2 role would work. Or reading them to a var before running the st2 role.
I see the logic introduced here does somewhat the same.

If we allow adding new vars for SSL cert file, we may allow the same for other vars https://github.com/StackStorm/ansible-st2#variables:
* `st2_config` -> `st2_config_file`
* `st2_rbac` -> `st2_rbac_file`
* `st2_ldap` -> `st2_ldap_file`
* `st2web_nginx_config` -> `st2web_nginx_config_file`
* `st2chatops_config` -> `st2chatops_config_file`

which looks like unnecessary complexity. With that, I think this kind of work could be done on the user's side.

Another problem with adding new vars named `st2web_ssl_certificate_path` and `st2web_ssl_certificate_key_path`, is that it might be confused with the nginx settings we already have here: https://github.com/StackStorm/st2/blob/abb694b85f75dc543ecea7df6b87def4bce53309/conf/nginx/st2.conf#L35-L36
```
  ssl_certificate           /etc/ssl/st2/st2.crt;
  ssl_certificate_key       /etc/ssl/st2/st2.key;
```
Eg. we're not changing the `/etc/ssl/st2/st2.crt`, but providing a path to read the certificate content on a remote server to later write it to the `/etc/ssl/st2/st2.crt`, which is a bit non-usual and might just bring more confusion from the user's side when using the role.

mamercad:
> Perhaps `cp` them on a remote host or similar as a prep task before running the st2 role would work. Or reading them to a var before running the st2 role. I see the logic introduced here does somewhat the same.
> 
> If we allow adding new vars for SSL cert file, we may allow the same for other vars https://github.com/StackStorm/ansible-st2#variables:
> 
> * `st2_config` -> `st2_config_file`
> * `st2_rbac` -> `st2_rbac_file`
> * `st2_ldap` -> `st2_ldap_file`
> * `st2web_nginx_config` -> `st2web_nginx_config_file`
> * `st2chatops_config` -> `st2chatops_config_file`
> 
> which looks like unnecessary complexity. With that, I think this kind of work could be done on the user's side.
> 
> Another problem with adding new vars named `st2web_ssl_certificate_path` and `st2web_ssl_certificate_key_path`, is that it might be confused with the nginx settings we already have here: https://github.com/StackStorm/st2/blob/abb694b85f75dc543ecea7df6b87def4bce53309/conf/nginx/st2.conf#L35-L36
> 
> ```
>   ssl_certificate           /etc/ssl/st2/st2.crt;
>   ssl_certificate_key       /etc/ssl/st2/st2.key;
> ```
> 
> Eg. we're not changing the `/etc/ssl/st2/st2.crt`, but providing a path to read the certificate content on a remote server to later write it to the `/etc/ssl/st2/st2.crt`, which is a bit non-usual and might just bring more confusion from the user's side when using the role.
> 
> After all, it's something that's left for maintainers to support. Any other ideas about the better solutions in terms of UX and role vars?

Yes; thanks for helping me through this -- I'm going to close this PR, I agree that adding the "alternate path way" isn't the very nice and likely would end up being more confusing than helpful.

Here's what I whipped up:

```yaml
---
- name: stackstorm
  hosts: stackstorm

  vars:

    st2web_ssl_certificate_path: /etc/ssl/certs/redacted-fullchain.crt
    st2web_ssl_certificate_key_path: /etc/ssl/private/redacted.pem

  tasks:

    - name: fetch our certificate key
      ansible.builtin.slurp:
        src: "{{ st2web_ssl_certificate_key_path }}"
      register: certificate_key
      become: true
      no_log: true

    - name: fetch our certificate
      ansible.builtin.slurp:
        src: "{{ st2web_ssl_certificate_path }}"
      register: certificate

    - name: store them in facts (named as stackstorm expect them to be named)
      ansible.builtin.set_fact:
        st2web_ssl_certificate_key: "{{ certificate_key.content | b64decode }}"
        st2web_ssl_certificate: "{{ certificate.content | b64decode }}"
      no_log: true

    - name: include stackstorm roles
      ansible.builtin.include_role:
        name: "{{ role }}"
      loop:
        - StackStorm.mongodb
        - StackStorm.rabbitmq
        - StackStorm.redis
        - StackStorm.st2repo
        - StackStorm.st2
        - StackStorm.nginx
        - StackStorm.st2web
        - StackStorm.nodejs
        - StackStorm.st2chatops
        - StackStorm.st2smoketests
      loop_control:
        loop_var: role
```

armab:
So the logic is getting a cert content into a var and passing it to the role.
Nice, thanks for sharing! :+1:

mamercad:
> So the logic is getting a cert content into a var and passing it to the role. Nice, thanks for sharing! 👍

Exactly; thanks again for the thoughtful feedback (this would have been a bad change to merge for sure).

PR #311, opened 2022-05-19
Allow Erlang version on EL8 to be pinned, and by default use < 25

amanda11:
Install on EL8 started to fail on RabbitMQ install. Problems reported "please re-compile this module with an Erlang/OTP 25 compiler"

Erlang 25 has been pushed to the RabbitMQ erlang repository, but according to https://www.rabbitmq.com/which-erlang.html erlang 25 support is preview.

Update so that erlang version can be specified, and make the default value 24* so do not pick up 25.

### DIFF
--- a/roles/StackStorm.rabbitmq/defaults/main.yml
+++ b/roles/StackStorm.rabbitmq/defaults/main.yml
@@ -6,4 +6,4 @@ rabbitmq_plugins: []
 # Set to "present" to install latest version, or specify specific version
 rabbitmq_version: "present"
 # Use version or wildcard
-erlang_el8_version: "24*"
+erlang_version: "{% if ansible_facts.os_family == 'Redhat' and ansible_facts.distribution_major_version == '8' %}24*{% else %}present{% endif %}"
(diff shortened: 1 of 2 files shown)


roles/StackStorm.rabbitmq/defaults/main.yml
```diff
@@ -5,3 +5,5 @@ rabbitmq_plugins: []
 #  - rabbitmq_management
 # Set to "present" to install latest version, or specify specific version
 rabbitmq_version: "present"
+# Use version or wildcard
+erlang_el8_version: "24*"
```
cognifloyd, line 9:
```suggestion
erlang_version: "24*"
```
It's not el8 specific, right?

amanda11: > It's not el8 specific, right?

Only EL8 failed last night. Looking at the packagecloud rabbit repo. EL8 has 25*, 24*, 23*. EL7 only has 23*.
So problem is EL8 specific at moment, and likely to not update EL7 as they never updated it for 24.

But you are right I at least need to do something as the file changed affects EL7 and EL8, but I'll have to have default versions for the two :-(

cognifloyd: In that case, you can add an extra `if` condition on the task that only uses the pinned version for el8

cognifloyd (approved):

PR #313, opened 2022-05-31
Use version of erlang < 1.25 and fix EL8 passlib

### LINKED ISSUE #312: Installation on EL8 reports failure due to passlib not installed
Since 29th May problems have been seen on installation with ansible and EL8. The ansible version in EPEL has moved to 5.x which installs python38.

Error reported:
```
       TASK [StackStorm.st2 : auth | Create htpasswd file] ****************************
       task path: /tmp/kitchen/roles/StackStorm.st2/tasks/auth.yml:34
       redirecting (type: modules) ansible.builtin.htpasswd to community.general.htpasswd
       An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ModuleNotFoundError: No module named 'passlib'
       fatal: [localhost]: FAILED! => {"changed": false, "msg": "Failed to import the required Python library (passlib) on a4b87b2336f8's Python /bin/python3.8. Please read the module documentation and install it in the appropriate location. If the required library is installed, but Ansible is using the wrong Python interpreter, please consult the documentation on ansible_python_interpreter"}
```

The problem is that the playbooks install python3-passlib which is a python 3.6 passlib module. But Ansible 5.x uses python3.8. There is no RPM built at moment for python3.8 passlib, see https://bugzilla.redhat.com/show_bug.cgi?id=2087268

amanda11:
Use the CloudSmith rabbitmq-erlang packages rather than the LaunchPad  erlang packages for Debian installations, so that we can specify a 24.x version of the rabbitmq-erlang. (25.x and 22.x are only available in LaunchPad for Debian and fail with latest rabbitMQ).

We no-longer want to use the latest 1.25 that is available in the rabbitmq-erlang repository, as this means that rabbitmq fails to install, and instead reports error:
```
fatal: [localhost]: FAILED! => {"attempts": 5, "cache_update_time": 1653987046, "cache_updated": false, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'rabbitmq-server'' failed: E: Unable to correct problems, you have held broken packages.\n", "rc": 100, "stderr": "E: Unable to correct problems, you have held broken packages.\n", "stderr_lines": ["E: Unable to correct problems, you have held broken packages."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nSome packages could not be installed. This may mean that you have\nrequested an impossible situation or if you are using the unstable\ndistribution that some required packages have not yet been created\nor been moved out of Incoming.\nThe following information may help to resolve the situation:\n\nThe following packages have unmet dependencies:\n rabbitmq-server : Depends: erlang-base (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            erlang-base-hipe (< 1:25.0) but it is not going to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-crypto (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-eldap (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-inets (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-mnesia (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-os-mon (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-parsetools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-public-key (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-runtime-tools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-ssl (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-syntax-tools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-tools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n                   Depends: erlang-xmerl (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or\n                            esl-erlang (< 1:25.0) but it is not installable\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Some packages could not be installed. This may mean that you have", "requested an impossible situation or if you are using the unstable", "distribution that some required packages have not yet been created", "or been moved out of Incoming.", "The following information may help to resolve the situation:", "", "The following packages have unmet dependencies:", " rabbitmq-server : Depends: erlang-base (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            erlang-base-hipe (< 1:25.0) but it is not going to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-crypto (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-eldap (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-inets (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-mnesia (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-os-mon (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-parsetools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-public-key (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-runtime-tools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-ssl (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-syntax-tools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-tools (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable", "                   Depends: erlang-xmerl (< 1:25.0) but 1:25.0-1rmq1ppa1~ubuntu18.04.1 is to be installed or", "                            esl-erlang (< 1:25.0) but it is not installable"]}
```

Resolves #312

### DIFF
--- a/roles/StackStorm.st2/tasks/auth.yml
+++ b/roles/StackStorm.st2/tasks/auth.yml
@@ -21,7 +21,7 @@
 - name: auth | Install auth pre-reqs (RedHat)
   become: yes
   yum:
-    name: 
+    name:
       - httpd-tools
       - "{{ passlib }}"
     state: present
--- a/roles/StackStorm.st2/tasks/auth.yml
+++ b/roles/StackStorm.st2/tasks/auth.yml
@@ -21,8 +21,7 @@
 - name: auth | Install auth pre-reqs (RedHat)
   become: yes
   yum:
-    name:
-      - httpd-tools
+    name: httpd-tools
     state: present
   register: _task
   retries: 5
@@ -35,8 +34,20 @@
   become: yes
   pip:
      name: passlib
+     executable: /usr/bin/pip3
      state: present
-  when: ansible_facts.os_family == 'RedHat'
+  when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version!='7'
+
+- name: auth | Install auth pre-reqs (RedHat)
+  become: yes
+  yum:
+    name: httpd-tools
+    state: present
+  register: _task
+  retries: 5
+  delay: 3
+  until: _task is succeeded
+  when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version=='7'
 
 - name: auth | Create htpasswd file
   become: true
--- a/roles/StackStorm.st2/tasks/auth.yml
+++ b/roles/StackStorm.st2/tasks/auth.yml
@@ -21,7 +21,9 @@
 - name: auth | Install auth pre-reqs (RedHat)
   become: yes
   yum:
-    name: httpd-tools
+    name: 
+      - httpd-tools
+      - "{{ passlib }}"
     state: present
   register: _task
   retries: 5
@@ -29,7 +31,8 @@
   until: _task is succeeded
   when: ansible_facts.os_family == 'RedHat'
 
-# Use pip to install passlib as no python38-passlib yum package available
+# Use pip to install passlib as no python38-passlib yum package available. Still need python3-passlib
+# installed as well though from above task
 - name: auth | Install pip pre-reqs (RedHat)
   become: yes
   pip:
@@ -37,17 +40,6 @@
      executable: /usr/bin/pip3
      state: present
   when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version!='7'
-
-- name: auth | Install auth pre-reqs (RedHat)
-  become: yes
-  yum:
-    name: "{{ passlib }}"
-    state: present
-  register: _task
-  retries: 5
-  delay: 3
-  until: _task is succeeded
-  when: ansible_facts.os_family == 'RedHat' and ansible_facts.distribution_major_version=='7'
 
 - name: auth | Create htpasswd file
   become: true
--- a/roles/StackStorm.st2/tasks/auth.yml
+++ b/roles/StackStorm.st2/tasks/auth.yml
@@ -23,12 +23,19 @@
   yum:
     name:
       - httpd-tools
-      - "{{ passlib }}"
     state: present
   register: _task
   retries: 5
   delay: 3
   until: _task is succeeded
+  when: ansible_facts.os_family == 'RedHat'
+
+# Use pip to install passlib as no python38-passlib yum package available
+- name: auth | Install pip pre-reqs (RedHat)
+  become: yes
+  pip:
+     name: passlib
+     state: present
   when: ansible_facts.os_family == 'RedHat'
 
 - name: auth | Create htpasswd file
--- a/roles/StackStorm.st2/tasks/auth.yml
+++ b/roles/StackStorm.st2/tasks/auth.yml
@@ -41,7 +41,7 @@
 - name: auth | Install auth pre-reqs (RedHat)
   become: yes
   yum:
-    name: httpd-tools
+    name: "{{ passlib }}"
     state: present
   register: _task
   retries: 5
--- a/roles/StackStorm.rabbitmq/defaults/main.yml
+++ b/roles/StackStorm.rabbitmq/defaults/main.yml
@@ -6,4 +6,4 @@ rabbitmq_plugins: []
 # Set to "present" to install latest version, or specify specific version
 rabbitmq_version: "present"
 # Use version or wildcard
-erlang_version: "{% if ansible_facts.os_family == 'Redhat' and ansible_facts.distribution_major_version == '8' %}24*{% elif ansible_facts.os_family == 'Debian' %}1:22.2.7+dfsg-1{% else %}present{% endif %}"
+erlang_version: "{% if ansible_facts.os_family == 'Redhat' and ansible_facts.distribution_major_version == '8' %}24*{% elif ansible_facts.os_family == 'Debian' %}1:24.3.4-1{% else %}present{% endif %}"
--- a/roles/StackStorm.rabbitmq/defaults/main.yml
+++ b/roles/StackStorm.rabbitmq/defaults/main.yml
@@ -6,4 +6,4 @@ rabbitmq_plugins: []
 # Set to "present" to install latest version, or specify specific version
 rabbitmq_version: "present"
 # Use version or wildcard
-erlang_version: "{% if ansible_facts.os_family == 'Redhat' and ansible_facts.distribution_major_version == '8' %}24*{% else %}present{% endif %}"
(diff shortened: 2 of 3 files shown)


amanda11:
Seems to be a problem with dependencies. I have been unable to install erlang and rabbitMQ compatible verisons on U20.

1) Previously we installed version 1:24.3.4-1rmq1ppa1~ubuntu20.04.1 when it was working successfully - for erlang-base etc. But this version doesn't seem to be available any more
2) The latest erlang-base etc version I can find is 1:22....
3) If you use the older erlang-base then you can't install the latest rabbitMQ as it requires erlang >= 1:23.2


So either:
1) we need to find the erlang-base, erlang-asn1 etc 1:24 versions we used before.  But I've not managed to find it.
2) I've found the erlang group package 1:24.3.4-1rmq1ppa1~ubuntu22.04.1 is available at https://launchpad.net/~rabbitmq/+archive/ubuntu/rabbitmq-erlang/+packages?field.name_filter=&field.status_filter=published&field.series_filter=
But if I attempt to install that it complains about missing dependencies - so that doesn't seem to work
3) If we go back to the earlier erlang as per this PR then we would have to rollback rabbitMQ version as well - and ideally we don't really want to rollback erlang version....

amanda11:
Using cloudsmith repos instead of LaunchPad allows us to install erlang version that works with rabbitmq
Using pip install of passlib on EL8 resolves the failing to find passlib for python3.8

roles/StackStorm.st2/tasks/auth.yml
```diff
@@ -31,6 +31,16 @@
   until: _task is succeeded
   when: ansible_facts.os_family == 'RedHat'
 
+# Use pip to install passlib as no python38-passlib yum package available. Still need python3-passlib
+# installed as well though from above task
+- name: auth | Install pip pre-reqs (RedHat)
+  become: yes
+  pip:
+     name: passlib
```
armab, line 39: Could we install the `passlib` via pip for all the OS flavors to get rid of this kind of mishmash:

https://github.com/StackStorm/ansible-st2/blob/02b5ffbde1ce35666bf0c106b80659270c9c6000/roles/StackStorm.st2/vars/debian.yml#L1-L3
https://github.com/StackStorm/ansible-st2/blob/02b5ffbde1ce35666bf0c106b80659270c9c6000/roles/StackStorm.st2/vars/redhat_8.yml#L1-L3
https://github.com/StackStorm/ansible-st2/blob/02b5ffbde1ce35666bf0c106b80659270c9c6000/roles/StackStorm.st2/vars/redhat.yml#L1-L3
https://github.com/StackStorm/ansible-st2/blob/02b5ffbde1ce35666bf0c106b80659270c9c6000/roles/StackStorm.st2/vars/debian_20.yml#L1-L3

amanda11: I think we'll still need to mess about a bit, as the pip executable differs. So we have to point to pip3 on EL8, but we'll need to point to different pip versions on some of the other installs. (I tried with just the default pip module and no executable name - but that failed :-( )
So we'll just be replacing the passlib variable with a pip variable.

armab (approved):
Thanks for the fixes! :+1:

roles/StackStorm.st2/tasks/auth.yml
amanda11: @armab We don't install default pip at moment for U20 as part of ansible install. So we'd need to add that in as well.
I'm not sure its going to make it any cleaner, just mean we'll have another pip on the box to perhaps get confused with....

armab: Either way looks like a rake.
Alright, thanks for checking.

LGTM :+1:

PR #317, opened 2022-06-23
Default `st2web_package_name: st2web`

### LINKED ISSUE #316: `st2web_package_name` is undefined
After pinning `st2web_version: 3.7.0-1`, I get the following error:

```
    The task includes an option with an undefined variable. The error was: 'st2web_package_name' is undefined
  
    The error appears to be in '/Users/mmercado/src/[redacted]/ansible-st2/roles/StackStorm.st2web/tasks/main.yml': line 26, column 3, but may
    be elsewhere in the file depending on the exact syntax problem.
  
    The offending line appears to be:
  
  
    - name: Install pinned st2web package
      ^ here
```

Initially, I had it unset (so, presumably defaulted to `stable`). On my second run, I decided to pin the version and hit the error above.

mamercad:
Fixes #316.

### DIFF
--- a/roles/StackStorm.st2web/defaults/main.yml
+++ b/roles/StackStorm.st2web/defaults/main.yml
@@ -12,3 +12,6 @@ st2web_ssl_certificate_key: null
 
 # String with a custom nginx configuration file to replace st2.conf. If not provided, the default st2.conf will be used.
 st2web_nginx_config: null
+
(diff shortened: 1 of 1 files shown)


amanda11 (approved):
Brilliant, thanks for the contribution. 
LGTM.

PR #321, opened 2022-09-29
Use latest erlang now by default - as erlang 25 now preferred erlang …

amanda11:
…version


armab (approved):
Thanks!

cognifloyd (approved):

mamercad:
Thanks for this, it fixed my deployments today 👋

PR #322, opened 2022-11-17
Need to install ansible-core on EL8

amanda11:
kitchen ansible provider always tries to install ansible, but that is no-longer available on EL8. 
Updated the el8 dockerfile to install ansible-core and the necessary collections that were in the old ansible package but not in ansible-core.


amanda11:
Installing ansible-core first, means that it gets passed the ansible error. However kitchen still tries to install ansible, as it fails on "which: command not found", when it checks if ansible is installed. But that error doesn't prevent build.
But it then fails at: 
ERROR! couldn't resolve module/action 'rabbitmq_plugin'. This often indicates a misspelling, missing collection, or incorrect module path.

cognifloyd (approved):

PR #323, opened 2022-12-04
Update and upgrade deprecated actions

mamercad:
Update the deprecated action `actions/setup-ruby@v1` to `ruby/setup-ruby@v1`. Upgrade `actions/checkout`, `actions/cache`, and `technote-space/workflow-conclusion-action` for the Node 12 (in favor of 16) deprecation. Looks like I was running into "rate limiting" ... I pushed `max-parallel: 1` as a shot in the dark.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -19,7 +19,7 @@ on:
 jobs:
   integration:
     name: 'integration - ${{matrix.name}}'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


cognifloyd:
I suspect you need to also switch to ubuntu-20.04 instead of ubuntu-latest. I had to do that with the main st2 repo because latest switched to ubuntu-22.04

mamercad:
> I suspect you need to also switch to `runs-on: ubuntu-20.04` instead of ubuntu-latest. I had to do that with the main st2 repo because latest switched to ubuntu-22.04

Ah yes, I've been burned by this elsewhere :loading:

.github/workflows/build.yaml
```diff
@@ -19,12 +19,13 @@ on:
 jobs:
   integration:
     name: 'integration - ${{matrix.name}}'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       fail-fast: false
+      max-parallel: 1
```
armab, line 25: cool stuff :+1:

PR #327, opened 2022-12-07
Refresh the `README.md`

mamercad:
Fix the broken Slack badge image, remove contractions, make grammar changes, write OS names and versions more consistently, show long options in commands, etc. I'll add markdown linting ~in another PR~ shortly.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -11,7 +11,7 @@ on:
     tags:
       - v*
   pull_request:
-    type: [opened, reopened, edited]
+    types: [opened, reopened, edited, synchronize]
   schedule:
     # run every night at midnight
     - cron:  '0 0 * * *'
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,6 +12,9 @@ on:
       - v*
   pull_request:
     types: [opened, reopened, edited, synchronize]
+    paths-ignore:
+      - .markdownlint.yml
+      - README.md
   schedule:
     # run every night at midnight
     - cron:  '0 0 * * *'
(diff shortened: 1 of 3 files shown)


armab (changes requested):
I'm :+1:  for the Readme refresh, but :-1: for Markdown linting as that's another uneeded check that might block a PR or change.

I think markdown linting makes sense for the projects where every file is documentation, but not for a single `README.md` when the frequency of change is low and the chance of breaking a Markdown file is even lower (they just work!).
Let's avoid utilizing computing resources in this case :green_apple:

mamercad:
> > I'm 👍 for the Readme refresh, but 👎 for Markdown linting as that's another uneeded check that might block a PR or change.
> > I think markdown linting makes sense for the projects where every file is documentation, but not for a single `README.md` when the frequency of change is low and the chance of breaking a Markdown file is even lower (they just work!). Let's avoid utilizing computing resources in this case 🍏
> 
> Sure, I'll push without the workflow shortly.

Done, I rebased as well so the history should be clean.

mamercad:
In the spirit of not running workflows unnecessarily, I'm noticing that the `build` workflow is running (for this `README.md` update, we should probably fix that).

.github/workflows/build.yaml
```diff
@@ -11,7 +11,10 @@ on:
     tags:
       - v*
   pull_request:
-    type: [opened, reopened, edited]
+    types: [opened, reopened, edited, synchronize]
+    paths-ignore:
+      - .markdownlint.yml
+      - README.md
```
armab, line 17: Nice! :+1:

armab (approved):
Thanks for the refresh!

Issue #242, opened 2019-05-17
Bump minimum version of Ansible to 2.7 or greater


cognifloyd:
Ansible 2.5 is EOL as of yesterday (2019-05-16) because :tada: Ansible ~2.8~ *2.9* is here.
2.8: https://groups.google.com/forum/#!topic/ansible-announce/OmkhL4zN1oo
2.9: https://groups.google.com/forum/#!topic/ansible-announce/chcjYohq6dA

I have not reviewed the ansible-st2 playbooks to see what we can do if we bump the version to 2.6.

2.6 has lots of bugfixes for include* and import*.
2.7 makes using vars with include* and import* more consistent.
2.8 adds support for content collections (repos like this one with more than one role). better jinja defaults if intermediate dict doesn't exist.
2.9 changes loop+var safety handling. ansible-galaxy collections support improvements

Current ansible [support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html) for 2.2 and above: 

Ansible Release | Latest Version | Status
-- | -- | --
devel | 2.10 (unreleased, trunk) | In development
2.9 | 2.9.2 (2019-12-04) | Supported (security and general bug fixes)
2.8 | 2.8.7 (2019-11-13) | Supported (security and critical bug fixes)
2.7 | 2.7.15 (2019-11-13) | Supported (security fixes)
2.6 | 2.6.20 (2019-10-17) | Unsupported (end of life)
2.5 | 2.5.15 (2019-02-21) | Unsupported (end of life)
2.4 | 2.4.6 (2018-07-05) | Unsupported (end of life)
2.3 | 2.3.3 (2017-12-20) | Unsupported (end of life)
2.2 | 2.2.3 (2017-05-09) | Unsupported (end of life)

2.9 is out so 2.6 is EOL.

armab:
Thanks @cognifloyd for the notice and putting it all together! 

We'll definitely need to do/plan migration to the next major Ansible version.

cognifloyd:
Chart updated. 2.9 is out. 2.6 is EOL.

setswei:
I would be happy to help with this. I am starting to implement stackstorm for my current employer and I am using Ansible to manage everything. i would be happy to assist with getting these roles up to the latest version

armab:
@setswei Awesome! Pull Requests are welcome!

setswei:
I will start working on this and get back to you guys soon.

amanda11:
[...]

PR #328, opened 2022-12-26
CI: Use Ansible Molecule for end-to-end testing

mamercad:
Switch to [Ansible Molecule](https://molecule.readthedocs.io/) for end-to-end testing in favor of [Kitchen Ansible](https://github.com/neillturner/kitchen-ansible) and [Kitchen Docker](https://github.com/test-kitchen/kitchen-docker) (this project is currently without a maintainer and has many known issues). Molecule is the de facto answer to Ansible and testing. This pull request covers convergence (and idempotence) using both Python 3.6 and Python 3.8 against StackStorm stable and unstable. The reason why there are two different jobs (`python36` and `python38`) is mainly to handle the different supported versions of Molecule. Lastly, this pull requests switches to completely running the end-to-end testing as non-root (which is nice since it validates that `become: true` works correctly); it appeared to me that the current end-to-end tests for Rocky Linux 8 and Focal used the `root` user (the same pattern for extending the existing `Dockerfiles` was used -- adding a non-`root` user named `molecule`). Here's an example running in my [fork](https://github.com/mamercad/ansible-st2/actions/runs/3782375120).


cognifloyd:
I haven't reviewed yet, but I love getting rid of our dependence on ruby here. It feels like such a huge context switch when I have to touch something with ruby and the ruby-ist APIs of kitchen sheet working exclusively in python and Ansible for awhile.

Yes please. Let's use molecule to test this instead of kitchen.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,142 @@
...
+    runs-on: ubuntu-20.04
+    strategy:
+      max-parallel: 1
+      fail-fast: false
+      matrix:
+        python:
+          - version: 3.6
+            ansible-core: 2.11.12
+            molecule: 3.6.1
+            molecule-docker: 1.1.0
```
armab, line 52: Screenshot 2022-12-29 at 12-21-41 CI Use Ansible Molecule for end-to-end testing by mamercad · Pull Request #328 · StackStorm_ansible-st2

The amount of versions here looks misleading, let's remove those and test only with one molecule version.
Similar to test-kitchen, as a developer/user who wants to modify or add Ansible tasks I don't care which version of Molecule we run as long as it runs the tests. Let's avoid overloading with this version info.

requirements.yml
```diff
@@ -0,0 +1,11 @@
...
+collections:
+  - name: ansible.posix
+    version: 1.4.0
+  - name: community.docker
+    version: 3.2.1
+  - name: community.general
+    version: 6.0.1
+  - name: community.rabbitmq
+    version: 1.2.3
+roles: []
```
armab, line 11: These are test requirements, right?
Could we move them under the `tests` dir?

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,142 @@
...
+          ansible-galaxy collection install --requirements-file requirements.yml
+
+      - name: Molecule test ${{ matrix.st2_repo.name }} under ${{ matrix.scenario.name }}
+        run: |
+          molecule test --scenario-name ${{ matrix.scenario.name }}
+        env:
+          HUBOT_SLACK_TOKEN: ${{ secrets.HUBOT_SLACK_TOKEN }}
+          ST2_REPO: ${{ matrix.st2_repo.name }}
+
+  python38:
```
armab, line 83: Interesting split matrix between python3.6 vs python3.8.

Thinking that once we add, change or remove the python version we'll just have more places to change in this repo which is undesired as we want to minimize the support burden.
From another side, highlighting py3.6 vs py3.8 could be useful from the quick-look standpoint.

I think on the Ansible side we care more about the OS in the first place (as before) and it's probably good to leave the python versioning matrix to the st2 repo.

armab (changes requested):
Thanks for working on this!

Test-Kitchen doesn't provide us any headaches, - it just works, but it's nice to switch to actively supported and developed Molecule and rely Ruby -> Python for this part.

----

It looks like the tests are failing because of how the Molecule runs the tests: iterating through the matrix for each Ansible task and it's falling on ChatOps tests due to rate-limiting.

I'd like to know if there's an option to split each check into a separate GH Job with a matrix and limit parallel runs with a factor of 2 or 3.
From that standpoint usability also matters as having a dedicated and clear check for each OS/platform is more helpful.
Screenshot 2022-12-29 at 12-36-07 CI Use Ansible Molecule for end-to-end testing by mamercad · Pull Request #328 · StackStorm_ansible-st2

mamercad:
> and it's falling on ChatOps tests due to rate-limiting.

Yeah, that's why I have it running serially, basically. I'll think on how to mitigate the rate-limiting even more.

.github/workflows/e2e.yaml
mamercad: I mean, one of the main reasons I chose to split the jobs between 3.6 and 3.8 was the Molecule version... would you rather just test both 3.6 and 3.8 against the same (lower) version of Molecule?

.github/workflows/e2e.yaml
mamercad: I'm not sure that I agree here, if we support 3.6 and 3.8, we should test them both.

requirements.yml
mamercad: I mean, yes and no, it depends on the Ansible version (i.e., before the split from "all-in-one" to Ansible core). In the newer world of `ansible-core` modules like [`rabbitmq_plugin`](https://github.com/StackStorm/ansible-st2/blob/b9673f50d89d48bd4e1ff0a9c64cd77bb47e2e0b/roles/StackStorm.rabbitmq/tasks/main.yml#L44) have been ripped out and place in the [collection](https://docs.ansible.com/ansible/latest/collections/community/rabbitmq/rabbitmq_plugin_module.html). But yes, I think we can put `requirements.yml` anywhere in this repository.

Beyond reworking testing (with Molecule) the next big change that this repository should undergo is that it should become an actual Ansible collection.

.github/workflows/e2e.yaml
armab: Yes, just having one pinned version of Molecule is fine.

.github/workflows/e2e.yaml
armab: We test python 3.6 and 3.8 in the https://github.com/stackStorm/st2 repository.

In the Ansible-st2 we care about the OS matrix (stable-el7, unstable-u20, etc).

armab:
> Yeah, that's why I have it running serially, basically. I'll think on how to mitigate the rate-limiting even more.

Please switch to the previous `{repo}-{os}` build matrix, there are several benefits behind that and it'll also help us with the rate-limiting naturally.

requirements.yml
armab: So sounds like `community.rabbitmq` is a core ansible-st2 requirement, how about others? 
Is the `community.docker` collection requirement for testing purposes only?

mamercad: Pushed without `community.docker`, that might have just been stray copy/paste.

.github/workflows/e2e.yaml
mamercad: > We test python 3.6 and 3.8 in the https://github.com/stackStorm/st2 repository.

You're not testing *Ansible* with Python 3.6 and 3.8 in that repository, are you?

mamercad: What I'm getting at is that the purpose of this repository is to ensure that *Ansible* can deploy StackStorm using Python 3.6 and Python 3.8, which is why, again, I think that we should *test* that.

armab: Let's re-iterate. I'd like to understand what we're trying to test, why, and the intent.

Based on the latest message, are you trying to test that Ansible could be ran in an environment that has python3.6 or python3.8 versions installed?
That host with Ansible can potentially setup remote StackStorm instance based on the roles we have here.

requirements.yml
armab: thanks, that looks good to me now as I see these are the core playbook requirements

mamercad:
> > Yeah, that's why I have it running serially, basically. I'll think on how to mitigate the rate-limiting even more.
> 
> Please switch to the previous `{repo}-{os}` build matrix, there are several benefits behind that and it'll also help us with the rate-limiting naturally.

Sure, do you care about testing both Python 3.6 and 3.8 *locally* (in CI) or do you want to just test 3.8?

.github/workflows/e2e.yaml
mamercad: > what we're trying to test, why, and the intent

What this PR is attempting to verify is that the Ansible roles and playbooks in this repository work in deploying StackStorm, using Ansible (with both Python 3.6 and 3.8 *locally*), to CentOS 7, Rocky Linux 8, Ubuntu 18, and Ubuntu 20. I think where we're crossing wires is that what I'm introducing here is validating the local combinations of Python 3.6 and 3.8 and their respective versions of Ansible Core. Again, what I think is important in regards to this repository is that it successfully uses Ansible and the playbooks and roles in this repository *locally* to deploy StackStorm *against* the supported operating systems. Since StackStorm supports Python 3.6 and 3.8 currently, I think it's important to test both of them *locally* in addition to *remotely*.

mamercad: In addition to the next big step of turning this repository into an Ansible collection, what we should do is be more formal in regards to the [requirements](https://github.com/StackStorm/ansible-st2#requirements). Having a matrix with *local* Python and Ansible combinations (as this PR introduces) will facilitate this, i.e., when a new version of Ansible Core is released, or when an older version of Ansible Core is no longer supported, it's simply a matter of bumping a couple of versions and all of a sudden we're testing on the new, and no longer testing on the old.

mamercad: > > > Yeah, that's why I have it running serially, basically. I'll think on how to mitigate the rate-limiting even more.
> > 
> > 
> > Please switch to the previous `{repo}-{os}` build matrix, there are several benefits behind that and it'll also help us with the rate-limiting naturally.
> 
> Sure, do you care about testing both Python 3.6 and 3.8 _locally_ (in CI) or do you want to just test 3.8?

[Here's](https://github.com/StackStorm/ansible-st2/pull/329) a draft PR which ignores 3.6 locally and matrixes across repo and OS, do you like this better?

armab:
I see. We don't care which python version is installed on the client(user) side that runs Ansible.
Ansible role repos usually don't test this.

From https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html#node-requirement-summary "Managed node Python" is more important, but it has very wide py version support between py2.7-py3.11. So I wouldn't think about this too as Ansible core already does a good job maintaining compatibility.


From the build matrix, what's important is the Ansible Version compatibility with the roles that we have (https://github.com/StackStorm/ansible-st2/issues/242), OS (el7, U20, etc), and repo (stable/unstable).

.github/workflows/e2e.yaml
amanda11: At a minimum you would expect our playbooks to be run on our target O/Ses to install StackStorm via ansible locally. So that users don't have to have a separate box to install on. Previously we were using random pythons in the ansible E2E as it was picking up various defaults, such as python3.9, python2.7 etc...

In that regard, then I think we'd need to be able to run with more than just python3.6 or python3.8 locally, as taking focal and bionic then they will have different python available.

armab: Per https://github.com/StackStorm/ansible-st2/pull/328#issuecomment-1369132002, Ansible client has a very wide py version support between `py2.7`-`py3.11`.

I think we're already in a good spot and can trust the py version compatibility advertised and tested by the Ansible project.

PR #329, opened 2022-12-31
CI: Use Ansible Molecule for end-to-end testing (Rework)

mamercad:
Relates to #328.


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,110 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python.version }}
+
+      - name: Install Python dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          python3 -m pip install --upgrade ansible-core==${{ matrix.python.ansible-core }}
+          python3 -m pip install --upgrade molecule==${{ matrix.python.molecule }}
+          python3 -m pip install --upgrade molecule-docker==${{ matrix.python.molecule-docker }}
```
armab, line 83: Is there a docker image with ansible (and maybe molecule?) already pre-installed?
as said, we can use any version of molecule, only the ansible core version is important in these tests.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,110 @@
...
+          # - version: 3.6
+          #   ansible-core: 2.11.12
+          #   molecule: 3.6.1
+          #   molecule-docker: 1.1.0
+          #   molecule-scenario: python36
+          - version: 3.8
+            ansible-core: 2.11.12
+            molecule: 3.6.1
+            molecule-docker: 1.1.0
+            molecule-scenario: python38
```
armab, line 67:
```suggestion
        ansible:
          - 2.11.12
```

Please include only the ansible core version in the build matrix.

Everything else could be hidden and pinned somewhere.
In other words, we don't care about `molecule`, `molecule-docker`, or `molecule-scenario` in the build matrix.
We even don't care if the e2e tests are run with a test-kitchen or molecule.

It shouldn't be exposed in the CI status string here to reduce the information overhead:
Screenshot 2023-01-02 at 19-10-07 CI Use Ansible Molecule for end-to-end testing (Rework) by mamercad · Pull Request #329 · StackStorm_ansible-st2

Adding an ansible version in there would be already an improvement, - thanks for that work. 
Screenshot 2023-01-02 at 19-12-06 Pull requests · StackStorm_ansible-st2
Let's just include only important info!

armab (changes requested):
Thanks for the changes!
Running a dedicated CI job per OS distro should be much better. 

Let's continue working on this approach :+1:

.github/workflows/e2e.yaml
mamercad: I'm not aware of one, I'll look around a bit.

.github/workflows/e2e.yaml
mamercad: Makes sense, I'll work on it.

.github/workflows/e2e.yaml
mamercad: Haven't had much luck finding one, in any event, setting up Python and the requirements takes about a minute.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,103 @@
...
+        st2_repo:
+          - stable
+          - unstable
+        os:
+          - centos7
+          - ubuntu18
+          - rockylinux8
+          - ubuntu20
+        ansible-core:
+          - "2.11.12"
```
armab, line 60: :+1:

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,103 @@
...
+
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v3
+
+      - name: Set up Python ${{ matrix.python.version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python.version }}
```
armab, line 69: Python is not in the matrix anymore, worth removing the `${{ matrix.python.version }}`

CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
+
+## In Development
+
+* Switch E2E testing to [Ansible Molecule](https://molecule.readthedocs.io/en/latest/). (by @mamercad)
```
armab, line 5: For simplicity/maintenance, we don't maintain the Changelog for this repo, but generate it for releases here:
https://github.com/StackStorm/ansible-st2/releases

So worth removing the file.

```suggestion
```

armab (changes requested):
Looks great!

Just left a couple more minor comments before merging.

.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,103 @@
+---
+name: E2E Tests
```
armab, line 2: I see we renaming the status name from `integration` -> `e2e tests`

Screenshot 2023-01-04 at 22-56-26 CI Use Ansible Molecule for end-to-end testing (Rework) by mamercad · Pull Request #329 · StackStorm_ansible-st2

Or we rename the [`tests/integration`](https://github.com/StackStorm/ansible-st2/tree/master/test/integration) dir or the `e2e tests` workflow.


I'm fine either way, as long as it's consistent in both places.

mamercad: As mentioned previously, once we turn this repository into a collection, the `tests/integration` directory will be needed, see [this](https://docs.ansible.com/ansible/latest/dev_guide/developing_collections_structure.html) for the structure of a collection. Though, the format of what's inside the `tests/integration` directory will change a bit, and, we'll want to start using `ansible-test` at that point in time (which doesn't necessarily obviate Molecule).

mamercad: I'm not sure what you mean by "consistent in both places", can you clarify?

armab (approved):
Looks good! Thanks for the changes :+1:

